PR #45, opened 2023-04-25
Add kuttl tests

jparoly:


tests/kuttl/tests/common/assert-sample-deployment.yaml
```diff
@@ -0,0 +1,41 @@
+apiVersion: designate.openstack.org/v1beta1
+kind: DesignateAPI
+metadata:
+  name: designate
+spec:
+  databaseInstance: openstack
+  databaseUser: designate
+  serviceUser: designate
+  containerImage: quay.io/tripleowallabycentos9/openstack-designate-api:current-tripleo
```
beagles, line 9: IIUC there is an initiative to switch to tcib images. Shouldn't we change the kuttl tests to use them as well? (see https://github.com/openstack-k8s-operators/designate-operator/pull/44)

jparoly: I don't see one on quay for designate.  Does it get created when it runs the first time?

jparoly: I updated this PR anyhow.

beagles (approved):
there seems to be on there now (https://quay.io/repository/podified-antelope-centos9/openstack-designate-api?tab=tags)

tests/kuttl/tests/common/assert-sample-deployment.yaml
jparoly: thanks, yeah i see it now, must have been more pages :-)

PR #50, opened 2023-05-22
Fix controller manager labeling

abays:
We now have a final form for operator controller-manager pod labeling of the format `openstack.org/operator-name: <operator name>` and are no longer relying on the inconsistent labels created by various `operator-sdk` versions.  All operators that lack this standardized pattern are being updated.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -12,10 +12,11 @@ metadata:
   namespace: system
   labels:
     control-plane: controller-manager
+    openstack.org/operator-name: designate
 spec:
   selector:
     matchLabels:
-      control-plane: controller-manager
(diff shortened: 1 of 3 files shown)


config/manager/manager.yaml
```diff
@@ -12,17 +12,19 @@ metadata:
   namespace: system
   labels:
     control-plane: controller-manager
+    openstack.org/operator-name: dataplane
 spec:
   selector:
     matchLabels:
-      control-plane: controller-manager
+      openstack.org/operator-name: dataplane
```
dprince, line 19: this should be designate, not dataplane right?

dprince (changes requested):

dkehn (approved):

PR #58, opened 2023-07-13
Fix defaulting of Replicas spec

kajinamit:
Currently the default values are ignored because these specs are set to 0 by webhook. Update the type from int32 to *int32 to avoid that behavior and ensure the expected default values are used.

This also adds common minumim/maximum limit about replicas.


dkehn (changes requested):
Looks good, but need to rebase and add changes to the other controllers - controllers/desiganteworker & mdns, etc.., e.g. https://github.com/openstack-k8s-operators/designate-operator/blob/main/controllers/designatemdns_controller.go#L493

kajinamit:
tbvh I'm not very confortable with the fact that some bugfix has been kept for a while without review while a feature is merged AFTER these bug fixes were proposed then immediately merged. And now you are requesting update after a while because of that order...

Anyway I'll push a new version to cover the current controllers.

dkehn (changes requested):
Can we get this propagated to all the v1bet1 type files and controllers and pkg deployments.go, thanks.

dkehn (approved):
LGTM

PR #59, opened 2023-07-13
Remove support for designate-agent

kajinamit:
The agent framework and drivers have already been deprecated[1].

[1] https://review.opendev.org/c/openstack/designate/+/872716


johnsom (approved):
LGTM, the agent is being removed upstream and we don't support any of the agent based drivers downstream.

PR #60, opened 2023-07-13
Remove endpoint url and keystone service id from status

kajinamit:
These items presented in CR status are not really used. Endpoints can be discovered using Keystone and we have no use case to require service id.


dkehn (commented):
can we get this adapted to all the controllers or is this only applicable to the api?

kajinamit:
> can we get this adapted to all the controllers or is this only applicable to the api?

This is irrelevant to the other controllers unless you need to create additional keystone endpoint/service.

dkehn (approved):
LGTM

PR #75, opened 2023-08-08
Change kuttl test to run in any namespace

vhariria:


fmount (approved):

kuttl-test.yaml
```diff
@@ -18,7 +18,6 @@ apiVersion: kuttl.dev/v1beta1
 kind: TestSuite
 reportFormat: JSON
 reportName: kuttl-test-designate
-namespace: openstack
```
cescgina, line 21: We should keep a namespace here as default, like `designate-kuttl-tests` so that we never run  the kuttl tests in a random namespace which won't work for openstack operators

vhariria: Thanks for your feedback. 
Pls clarify, as we previously kept a custom namespace in operators such as keystone or glance but not in a recent [ octavia](https://github.com/openstack-k8s-operators/octavia-operator/pull/165/files) PR.

cescgina: It's there, it was simply moved within the file, see:https://github.com/openstack-k8s-operators/octavia-operator/pull/165/files#diff-b86f32129f2f798ed8705ac65814aca39658c01e41f36d3777bfb17e0185452bR19

vhariria: I see, will do.

dkehn (approved):
LGTM

PR #77, opened 2023-08-17
Fix a few missing CRs in csv

kajinamit:
CSV does not contain some CRs recently added. This ensures all CRs are listed.


abays (approved):
/lgtm

PR #81, opened 2023-08-24
Stop using KOLLA_CONFIG_FILE environment

kajinamit:
and refactors how volumes and volume mounts are managed.

I also noticed this operator currently lacks the logic to present customServiceConfig in sub CR layer but that is not covered by this change and should be addressed separately.


dkehn (approved):
LGTM

PR #82, opened 2023-08-25
Remove KOLLA_CONFIG_FILE environment variable

kajinamit:
The previous attempt added the bind mount to expose the config file to the default path but it did not remove the environment variable actually from pods.


dkehn (approved):
LGTM

PR #102, opened 2023-10-02
Adding some health checks

beagles:
In some cases the health checks will need to be improved beyond "is there a process with the expected name".


dkehn (approved):
LGTM

PR #103, opened 2023-10-02
Fix transposed characters for designate

beagles:
There are a few locations where designate is mispelled that might cause confusion or configuration errors.


dkehn (approved):
LGTM

PR #129, opened 2023-11-21
Support disconnected network environments

dprince:
This PR adds support for installing the operator in disconnected network environments. To build with image-digests set USE_IMAGE_DIGESTS=true before running make bundle.

For Prow jobs we are enabling this via .prow-ci.env

(NOTE: this currently requires a secure registry)


dkehn (approved):
LGTM

PR #136, opened 2023-12-13
update Makefile to include local bin operator-sdk

yazug:
specify version for operator-sdk for make bundle
download operator-sdk pinned version when generating bundle update make bundle to use local operator-sdk
update to 1.31 version

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -179,10 +179,12 @@ $(LOCALBIN):
 KUSTOMIZE ?= $(LOCALBIN)/kustomize
 CONTROLLER_GEN ?= $(LOCALBIN)/controller-gen
 ENVTEST ?= $(LOCALBIN)/setup-envtest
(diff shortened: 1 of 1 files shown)


config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/openstack-k8s-operators/designate-operator
-  newTag: latest
+  newName: registry-proxy.engineering.redhat.com/rh-osbs/rhosp-dev-preview-trunk-designate-rhel9-operator
```
viroel, line 15: @yazug , is this needed?

yazug: no not needed, reverted that change and rebased to current. Thanks for spotting that.

abays (approved):
/lgtm

PR #156, opened 2024-02-19
migrate from databaseUsername to databaseAccount and fully use MariaDBAccount

zzzeek:
This moves designate to fully use MariaDBAccount based on the dev work being done for mariadb-operator:

Lead Jira: [OSPRH-4095](https://issues.redhat.com/browse/OSPRH-4095)

1. [x] controller calls EnsureMariaDBAccount up front to make sure MariaDBAccount is present
2. [x] error code from EnsureMariaDBAccount is handled, Conditions are amended when error is returned
3. [x] controller calls NewDatabaseForAccount instead of NewDatabase
4. [x]  GetAccountAndSecret is used to retrieve account /secret to populate template
5. [x]  GetDatabaseByName() , normally used for delete finalizers, replaced with GetDatabaseByNameAndAccount
6. [x]  CreateOrPatchAll() used to patch the Database, replacing CreateOrPatchDB / CreateOrPatchDBByName
7. [x]  controller calls DeleteUnusedMariaDBAccountFinalizers when launched pods are definitely running on a new MariaDBAccount, returns error code if present
8. [x]  PasswordSelectors that refer to database are removed
9. [x]  all databaseUser replaced with databaseAccount inside of all XYZ_types.go
10. [x] all databaseUser replaced with databaseAccount inside of all `kuttl/*.yaml`
11. [x] all default databaseUser names become databaseAccount, replacing underscores with dashes inside of all XYZ_types.go
12. [x] all default databaseUser names become databaseAccount, replacing underscores with dashes inside of all `kuttl/*.yaml`
13. [ ] ~~MariaDBAccountSuiteTests are used in controller ginkgo tests if it has them~~ N/A , no ginkgo suite
14. [x] Use configsecrets for database URLs; remove from job hash
15. [x] [184](https://github.com/openstack-k8s-operators/mariadb-operator/pull/184) is merged and replaces from go.mod are removed

### DIFF
--- a/controllers/designate_controller.go
+++ b/controllers/designate_controller.go
@@ -38,7 +38,6 @@ import (
 	keystonev1 "github.com/openstack-k8s-operators/keystone-operator/api/v1beta1"
 	"github.com/openstack-k8s-operators/lib-common/modules/common"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/condition"
-	"github.com/openstack-k8s-operators/lib-common/modules/common/configmap"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/endpoint"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/env"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/helper"
@@ -295,7 +294,7 @@ func (r *DesignateReconciler) reconcileDelete(ctx context.Context, instance *des
 	Log.Info(fmt.Sprintf("Reconciling Service '%s' delete", instance.Name))
 
 	// remove db finalizer first
-	db, err := mariadbv1.GetDatabaseByName(ctx, helper, instance.Name)
+	db, err := mariadbv1.GetDatabaseByNameAndAccount(ctx, helper, instance.Name, instance.Spec.DatabaseAccount, instance.Namespace)
 	if err != nil && !k8s_errors.IsNotFound(err) {
 		return ctrl.Result{}, err
 	}
@@ -330,19 +329,16 @@ func (r *DesignateReconciler) reconcileInit(
 	//
 	// create service DB instance
 	//
-	db := mariadbv1.NewDatabase(
-		instance.Name,
-		instance.Spec.DatabaseUser,
-		instance.Spec.Secret,
-		map[string]string{
-			"dbName": instance.Spec.DatabaseInstance,
-		},
+	db := mariadbv1.NewDatabaseForAccount(
+		instance.Spec.DatabaseInstance, // mariadb/galera service to target
+		instance.Name,                  // name used in CREATE DATABASE in mariadb
+		instance.Name,                  // CR name for MariaDBDatabase
+		instance.Spec.DatabaseAccount,  // CR name for MariaDBAccount
+		instance.Namespace,             // namespace
 	)
+
 	// create or patch the DB
-	ctrlResult, err := db.CreateOrPatchDB(
-		ctx,
-		helper,
-	)
+	ctrlResult, err := db.CreateOrPatchAll(ctx, helper)
 	if err != nil {
 		instance.Status.Conditions.Set(condition.FalseCondition(
 			condition.DBReadyCondition,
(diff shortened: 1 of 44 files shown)


templates/designate/config/designate.conf
```diff
@@ -20,11 +20,12 @@ healthcheck_enabled=True
 connection = mysql+pymysql://${DBUSER}:${DBPASSWORD}@${DBHOST}/${DB}
 
 [storage:sqlalchemy]
-connection = mysql+pymysql://root:${DBPASSWORD}@${DBHOST}/${DB}?charset=utf8
```
zzzeek, line 23: OK so this one is quite strange, and the use of "root" for the username seems intentional here.

So we would guess there's some anomalous behavior when running migrations (still using sqlalchemy-migrate??) or later on where somehow the regular DB user doesnt have enough privs.   which would be very unusual for mysql/mariadb,  this is a straightforward kind of DB for which we give full grants to the user and if there's something that doesnt work with the regular user, we should fix that (and I least need to know what it is).

But if this *does* need to be root then we cant use `${DBPASSWORD}` here, we would need to add a new token `${DBROOTPASSWORD}` and then pass that in from osp-secret under the `DatabasePassword` selector, but even then this is kind of not really planned in the overall architecture and we really need this to be a non-root user.

controllers/designate_controller.go
```diff
@@ -574,6 +579,17 @@ func (r *DesignateReconciler) reconcileNormal(ctx context.Context, instance *des
...
+	// openstack-operator or the cloud operator up front without a specific
+	// MariaDBDatabase configured yet.   Otherwise, a MariaDBAccount CR is
+	// created here with a generated username as well as a secret with
+	// generated password.   The MariaDBAccount is created without being
+	// yet associated with any MariaDBDatabase.
+	_, _, err = mariadbv1.EnsureMariaDBAccount(
+		ctx, helper, instance.Spec.DatabaseAccount,
+		instance.Namespace, false,
+	)
+
```
zzzeek, line 592: need to add a status update here, and also return if err is non nil, example https://github.com/openstack-k8s-operators/neutron-operator/blame/13c5c990d917ab2825bda8e44b6f631474cc9fe1/controllers/neutronapi_controller.go#L925

pkg/designate/dbsync.go
```diff
@@ -78,10 +80,10 @@ func DbSyncJob(
 	initContainerDetails := APIDetails{
 		ContainerImage:       instance.Spec.DesignateAPI.ContainerImage,
 		DatabaseHost:         instance.Status.DatabaseHostname,
-		DatabaseUser:         instance.Spec.DatabaseUser,
```
zzzeek, line 81: OK this likely has to change because username and secret name are now dynamic, they change.   The job hash here cannot be determined with username/secret name.   Heat controller has the same problem.   somehow other contrllers like keystone etc. do not need to pass DB username/secret here and i am hoping this does not imply some huge change to kolla config or something

templates/designate/config/designate.conf
zzzeek: im 99% sure this was just some remnant of something from ancient history, also the utf8 part we dont need since we create the database with utf8 encoding by default in any case, the clients we use these days are generally utf8 oriented overall ever since Python 3 made Unicode mandatory.

templates/designate/config/designate.conf
```diff
@@ -17,14 +17,15 @@ transport_url=rabbit://stackrabbit:secret@10.0.110.9:5672/
...
 [database]
-connection = mysql+pymysql://${DBUSER}:${DBPASSWORD}@${DBHOST}/${DB}
+connection={{ .DatabaseConnection }}
 
 [storage:sqlalchemy]
-connection = mysql+pymysql://root:${DBPASSWORD}@${DBHOST}/${DB}?charset=utf8
+connection={{ .DatabaseConnection }}
 
 [coordination]
 backend_url=memcached://127.0.0.1:11211
```
stuggi, line 26: unrelated to this PR, but that looks not the correct  memcached?

stuggi (approved):
/lgtm

PR #161, opened 2024-04-09
Add ObservedGeneration to the sub Resources

fmount:
This patch does a few things:
1. it adds `observedGeneration` to the sub custom resources
2. it proposes to bump the `observedGeneration` at the beginning of the reconciliation loop (knative/serving#4937)
3. it checks, at the top level, if the `ObservedGeneration` matches with the metadata.generation assigned to the subCR(s)
4. before marking the `DeploymentReadyCondition` as `True`, the `ObservedGeneration` is compared with the `Generation` of the `Deployment`

### DIFF
--- a/controllers/designate_controller.go
+++ b/controllers/designate_controller.go
@@ -197,7 +197,6 @@ func (r *DesignateReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		condition.UnknownCondition(designatev1beta1.DesignateMdnsReadyCondition, condition.InitReason, designatev1beta1.DesignateMdnsReadyInitMessage),
 		condition.UnknownCondition(designatev1beta1.DesignateProducerReadyCondition, condition.InitReason, designatev1beta1.DesignateProducerReadyInitMessage),
 		condition.UnknownCondition(designatev1beta1.DesignateBackendbind9ReadyCondition, condition.InitReason, designatev1beta1.DesignateBackendbind9ReadyInitMessage),
-		condition.UnknownCondition(condition.DeploymentReadyCondition, condition.InitReason, condition.DeploymentReadyInitMessage),
 		condition.UnknownCondition(condition.NetworkAttachmentsReadyCondition, condition.InitReason, condition.NetworkAttachmentsReadyInitMessage),
 		// service account, role, rolebinding conditions
 		condition.UnknownCondition(condition.ServiceAccountReadyCondition, condition.InitReason, condition.ServiceAccountReadyInitMessage),
@@ -1409,15 +1408,15 @@ func (r *DesignateReconciler) checkDesignateWorkerGeneration(
 	instance *designatev1beta1.Designate,
 ) (bool, error) {
 	Log := r.GetLogger(context.Background())
-	central := &designatev1beta1.DesignateWorkerList{}
+	worker := &designatev1beta1.DesignateWorkerList{}
 	listOpts := []client.ListOption{
 		client.InNamespace(instance.Namespace),
 	}
-	if err := r.Client.List(context.Background(), central, listOpts...); err != nil {
+	if err := r.Client.List(context.Background(), worker, listOpts...); err != nil {
 		Log.Error(err, "Unable to retrieve DesignateWorker %w")
 		return false, err
 	}
-	for _, item := range central.Items {
+	for _, item := range worker.Items {
 		if item.Generation != item.Status.ObservedGeneration {
 			return false, nil
 		}
--- a/controllers/designate_controller.go
+++ b/controllers/designate_controller.go
@@ -204,8 +204,8 @@ func (r *DesignateReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		condition.UnknownCondition(condition.RoleReadyCondition, condition.InitReason, condition.RoleReadyInitMessage),
 		condition.UnknownCondition(condition.RoleBindingReadyCondition, condition.InitReason, condition.RoleBindingReadyInitMessage),
 	)
-
 	instance.Status.Conditions.Init(&cl)
+	instance.Status.ObservedGeneration = instance.Generation
 
 	// If we're not deleting this and the service object doesn't have our finalizer, add it.
 	if instance.DeletionTimestamp.IsZero() && controllerutil.AddFinalizer(instance, helper.GetFinalizer()) || isNewInstance {
@@ -680,17 +680,34 @@ func (r *DesignateReconciler) reconcileNormal(ctx context.Context, instance *des
 			err.Error()))
 		return ctrl.Result{}, err
 	}
-	if op != controllerutil.OperationResultNone {
-		Log.Info(fmt.Sprintf("Deployment %s successfully reconciled - operation: %s", instance.Name, string(op)))
+	apiObsGen, err := r.checkDesignateAPIGeneration(instance)
+	if err != nil {
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			designatev1beta1.DesignateAPIReadyCondition,
+			condition.ErrorReason,
+			condition.SeverityWarning,
+			designatev1beta1.DesignateAPIReadyErrorMessage,
+			err.Error()))
+		return ctrlResult, nil
+	}
+	if !apiObsGen {
+		instance.Status.Conditions.Set(condition.UnknownCondition(
+			designatev1beta1.DesignateAPIReadyCondition,
+			condition.InitReason,
+			designatev1beta1.DesignateAPIReadyInitMessage,
+		))
+	} else {
+		// Mirror DesignateAPI status' ReadyCount to this parent CR
+		instance.Status.DesignateAPIReadyCount = designateAPI.Status.ReadyCount
+		// Mirror DesignateAPI's condition status
+		c := designateAPI.Status.Conditions.Mirror(designatev1beta1.DesignateAPIReadyCondition)
+		if c != nil {
+			instance.Status.Conditions.Set(c)
+		}
 	}
 
-	// Mirror DesignateAPI status' ReadyCount to this parent CR
-	instance.Status.DesignateAPIReadyCount = designateAPI.Status.ReadyCount
-
-	// Mirror DesignateAPI's condition status
-	c := designateAPI.Status.Conditions.Mirror(designatev1beta1.DesignateAPIReadyCondition)
-	if c != nil {
-		instance.Status.Conditions.Set(c)
+	if op != controllerutil.OperationResultNone && apiObsGen {
+		Log.Info(fmt.Sprintf("Deployment %s successfully reconciled - operation: %s", instance.Name, string(op)))
 	}
 	Log.Info("Deployment API task reconciled")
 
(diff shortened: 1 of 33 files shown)


controllers/designate_controller.go
```diff
@@ -805,7 +890,24 @@ func (r *DesignateReconciler) reconcileNormal(ctx context.Context, instance *des
...
+			err.Error()))
+		return ctrlResult, nil
+	}
+	if !bindObsGen {
+		instance.Status.Conditions.Set(condition.UnknownCondition(
+			designatev1beta1.DesignateBackendbind9ReadyCondition,
+			condition.InitReason,
+			designatev1beta1.DesignateBackendbind9ReadyInitMessage,
+		))
+	}
```
gibizer, line 909: ditto about mirroring

controllers/designate_controller.go
```diff
@@ -830,7 +932,24 @@ func (r *DesignateReconciler) reconcileNormal(ctx context.Context, instance *des
...
+			err.Error()))
+		return ctrlResult, nil
+	}
+	if !unbObsGen {
+		instance.Status.Conditions.Set(condition.UnknownCondition(
+			designatev1beta1.DesignateUnboundReadyCondition,
+			condition.InitReason,
+			designatev1beta1.DesignateUnboundReadyInitMessage,
+		))
+	}
```
gibizer, line 951: ditto about mirroring

controllers/designateapi_controller.go
```diff
@@ -760,7 +761,8 @@ func (r *DesignateAPIReconciler) reconcileNormal(ctx context.Context, instance *
 		return ctrl.Result{}, err
 	}
 
-	if instance.Status.ReadyCount > 0 {
+	if instance.Status.ReadyCount > 0 &&
```
gibizer, line 764: the ReadyCount > 0 statement could cause trouble during update. E.g. if there are 3 replicas requested and one is ready with the new image it does not mean that the other two replicas are running the new image as well.

controllers/designatebackendbind9_controller.go
```diff
@@ -565,7 +566,8 @@ func (r *DesignateBackendbind9Reconciler) reconcileNormal(ctx context.Context, i
 		return ctrl.Result{}, err
 	}
 
-	if instance.Status.ReadyCount > 0 {
+	if instance.Status.ReadyCount > 0 &&
```
gibizer, line 569: ditto

controllers/designatecentral_controller.go
```diff
@@ -523,7 +524,8 @@ func (r *DesignateCentralReconciler) reconcileNormal(ctx context.Context, instan
 		return ctrl.Result{}, err
 	}
 
-	if instance.Status.ReadyCount > 0 {
+	if instance.Status.ReadyCount > 0 &&
```
gibizer, line 527: ditto

controllers/designatemdns_controller.go
```diff
@@ -521,7 +522,8 @@ func (r *DesignateMdnsReconciler) reconcileNormal(ctx context.Context, instance
 		return ctrl.Result{}, err
 	}
 
-	if instance.Status.ReadyCount > 0 {
+	if instance.Status.ReadyCount > 0 &&
```
gibizer, line 525: ditto

controllers/designateproducer_controller.go
```diff
@@ -523,7 +524,8 @@ func (r *DesignateProducerReconciler) reconcileNormal(ctx context.Context, insta
 		return ctrl.Result{}, err
 	}
 
-	if instance.Status.ReadyCount > 0 {
+	if instance.Status.ReadyCount > 0 &&
```
gibizer, line 527: ditto

controllers/designateworker_controller.go
```diff
@@ -518,7 +519,8 @@ func (r *DesignateWorkerReconciler) reconcileNormal(ctx context.Context, instanc
 		return ctrl.Result{}, err
 	}
 
-	if instance.Status.ReadyCount > 0 {
```
gibizer, line 521: ditto

controllers/designateunbound_controller.go
```diff
@@ -125,6 +125,7 @@ func (r *UnboundReconciler) Reconcile(ctx context.Context, req ctrl.Request) (re
 	)
 
 	instance.Status.Conditions.Init(&cl)
+	instance.Status.ObservedGeneration = instance.Generation
```
gibizer, line 128: you are missed the generation check for the deployment created by this controller. https://github.com/openstack-k8s-operators/designate-operator/blob/29d27af2376793a54cde1cd9abbb867d71ac1290/controllers/designateunbound_controller.go#L379-L382

gibizer (commented):
the ReadyCount > 0 can be fixed in a follow up

controllers/designate_controller.go
```diff
@@ -755,7 +806,24 @@ func (r *DesignateReconciler) reconcileNormal(ctx context.Context, instance *des
...
+	if err != nil {
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			designatev1beta1.DesignateMdnsReadyCondition,
+			condition.ErrorReason,
+			condition.SeverityWarning,
+			designatev1beta1.DesignateMdnsReadyErrorMessage,
+			err.Error()))
+		return ctrlResult, nil
+	}
+	if !workerObsGen {
```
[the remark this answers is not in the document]
fmount: Fixing

controllers/designate_controller.go
```diff
@@ -780,7 +848,24 @@ func (r *DesignateReconciler) reconcileNormal(ctx context.Context, instance *des
...
+	if err != nil {
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			designatev1beta1.DesignateProducerReadyCondition,
+			condition.ErrorReason,
+			condition.SeverityWarning,
+			designatev1beta1.DesignateProducerReadyErrorMessage,
+			err.Error()))
+		return ctrlResult, nil
+	}
+	if !workerObsGen {
```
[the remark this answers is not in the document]
fmount: Fixing

controllers/designateapi_controller.go
fmount: Yeah, I think we had  a similar conversation for `octavia-operator`. Fixing it and moving it to `instance.Status.ReadyCount == *instance.Spec.Replicas`, to make sure we report readiness only after the Pods have been rolled out.

controllers/designateunbound_controller.go
fmount: good catch, thank you!

fmount: Done

controllers/designateapi_controller.go
fmount: Done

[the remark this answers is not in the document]
fmount: not sure if this is the same comment you had before or you are referring to something else: I think that we should compare `instance.Status.ReadyCount` w/ `instance.Spec.Replicas` to make sure we mark the condition as True only if we rolled out all the desired Pods.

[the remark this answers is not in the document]
gibizer: I think this is the same comment. We can fix this in a followup

controllers/designateapi_controller.go
```diff
@@ -725,43 +726,45 @@ func (r *DesignateAPIReconciler) reconcileNormal(ctx context.Context, instance *
...
-		networkReady, networkAttachmentStatus, err = nad.VerifyNetworkStatusFromAnnotation(
-			ctx,
-			helper,
-			instance.Spec.NetworkAttachments,
-			serviceLabels,
-			instance.Status.ReadyCount,
-		)
-		if err != nil {
-			return ctrl.Result{}, err
+	if depl.GetDeployment().Generation <= depl.GetDeployment().Status.ObservedGeneration {
```
gibizer, line 729: based on the discussion in the docs PR https://github.com/openstack-k8s-operators/dev-docs/pull/102#discussion_r1565989013 I think the < case also means we need to reconcile again instead of declaring success.j

controllers/designatecentral_controller.go
```diff
@@ -488,43 +489,46 @@ func (r *DesignateCentralReconciler) reconcileNormal(ctx context.Context, instan
...
-			ctx,
-			helper,
-			instance.Spec.NetworkAttachments,
-			serviceLabels,
-			instance.Status.ReadyCount,
-		)
-		if err != nil {
-			return ctrl.Result{}, err
+
+	if depl.GetDeployment().Generation <= depl.GetDeployment().Status.ObservedGeneration {
```
gibizer, line 493: I would not declare success on the < case

controllers/designatemdns_controller.go
```diff
@@ -486,43 +487,46 @@ func (r *DesignateMdnsReconciler) reconcileNormal(ctx context.Context, instance
...
-			ctx,
-			helper,
-			instance.Spec.NetworkAttachments,
-			serviceLabels,
-			instance.Status.ReadyCount,
-		)
-		if err != nil {
-			return ctrl.Result{}, err
+
+	if depl.GetDeployment().Generation <= depl.GetDeployment().Status.ObservedGeneration {
```
gibizer, line 491: ditto

controllers/designateproducer_controller.go
```diff
@@ -488,43 +489,46 @@ func (r *DesignateProducerReconciler) reconcileNormal(ctx context.Context, insta
...
-			ctx,
-			helper,
-			instance.Spec.NetworkAttachments,
-			serviceLabels,
-			instance.Status.ReadyCount,
-		)
-		if err != nil {
-			return ctrl.Result{}, err
+
+	if depl.GetDeployment().Generation <= depl.GetDeployment().Status.ObservedGeneration {
```
gibizer, line 493: ditto

controllers/designateunbound_controller.go
```diff
@@ -342,43 +343,45 @@ func (r *UnboundReconciler) reconcileNormal(ctx context.Context, instance *desig
...
-		networkReady, networkAttachmentStatus, err = nad.VerifyNetworkStatusFromAnnotation(
-			ctx,
-			helper,
-			instance.Spec.NetworkAttachments,
-			serviceLabels,
-			instance.Status.ReadyCount,
-		)
-		if err != nil {
-			return ctrl.Result{}, err
+	if depl.GetDeployment().Generation <= depl.GetDeployment().Status.ObservedGeneration {
```
gibizer, line 346: ditto

controllers/designatecentral_controller.go
fmount: We're not returning success, we're just proceeding in the reconciliation loop. I'm not sure we need to explicitly handle the `<` case and return.
In theory we can continue the reconciliation loop if `<=`, and not stop. The deployement controller is supposed to handle any inconsistencies before it returns Ready to our controller.

controllers/designatemdns_controller.go
fmount: see comment above

controllers/designateproducer_controller.go
fmount: see comment above

controllers/designateunbound_controller.go
fmount: see comment above

controllers/designatecentral_controller.go
fmount: Otherwise let's go back to `==` condition and let the current reconciliation return false w/o the need of necessarily handling the `<` case

gibizer: Sorry I wasn't precise enough. By allowing to consider the ReadyReplicas in the case of < the controller might declare success on a status that has a newer generation than what this controller requested. If that spec generation introduced a change that this controller does not agree with then the controller should not accept the status of the reconciliation of the spec version.

> The deployement controller is supposed to handle any inconsistencies before it returns Ready to our controller.

the deployment controller will return ready for a spec generation that is not set by this controller so we should ignore that readyness until we can make sure that such spec generation does not contradict with the spec this controller wants.

> Otherwise let's go back to == condition and let the current reconciliation return false w/o the need of necessarily handling the < case

That sounds good to me.

gibizer (approved):
looks good.

controllers/designate_controller.go
```diff
@@ -680,6 +680,23 @@ func (r *DesignateReconciler) reconcileNormal(ctx context.Context, instance *des
...
+			err.Error()))
+		return ctrlResult, nil
+	}
+	if !apiObsGen {
+		instance.Status.Conditions.Set(condition.UnknownCondition(
+			designatev1beta1.DesignateAPIReadyCondition,
+			condition.InitReason,
+			designatev1beta1.DesignateAPIReadyInitMessage,
+		))
+	}
```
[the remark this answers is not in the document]
fmount: Done

controllers/designatecentral_controller.go
fmount: Done, closing the thread as we switched to `==`

fmount:
> the ReadyCount > 0 can be fixed in a follow up

I think this is covered as well.

fmount:
@dprince FYI I tested this locally, when CI is green I think is good to go

fmount:
@gibizer @dprince wondering if we can merge this

gibizer (commented):
still looks good

gibizer (approved):
looks like a clean rebase so this is still good to go

abays (approved):
/lgtm

PR #163, opened 2024-04-23
Ensure golangci-lint runs on all files

gibizer:
* removed unused func params
* ignore revive dot-imports rule on ginkgo and gomega as dot import
there is the recommended practice
* various small style fixes found by the linter
* the linter also showed that we the network attachments are unused. This was probably missed during some refactor as the network attachments was gathered but isn't passed anywhere. Based on the controller logic we can pass it to the dbsync job.


abays (approved):
/lgtm

PR #164, opened 2024-04-25
Remove explicit Update on instance

gibizer:
The unbound controller did a separate client.Update after
it added the self finalizer and then continued the reconciliation
However the client.Update only wrote the
main resource and ignored the status subresource. However it refreshed
the whole instance with the result of the Update meaning it reset the in
memory status updates. We don't need two separate update so this PR
removes the standalone Update and return immediately to rely on the
deferred PatchInstance call to do all the updates. Similar logic is
applied to the delete code path.

### DIFF
--- a/controllers/designateunbound_controller.go
+++ b/controllers/designateunbound_controller.go
@@ -161,10 +161,6 @@ func (r *UnboundReconciler) reconcileDelete(ctx context.Context, instance *desig
 	util.LogForObject(helper, "Reconciling Service delete", instance)
 	controllerutil.RemoveFinalizer(instance, helper.GetFinalizer())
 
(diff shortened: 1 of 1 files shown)


abays (approved):
/lgtm

abays:
```
{  "designate-operator-build-deploy-kuttl" test steps failed: "designate-operator-build-deploy-kuttl" pod "designate-operator-build-deploy-kuttl-openstack-k8s-operators-deploy" failed: could not watch pod: the pod ci-op-2mjq3897/designate-operator-build-deploy-kuttl-openstack-k8s-
```

/test designate-operator-build-deploy-kuttl

PR #187, opened 2024-07-05
Use MariaDB account and secret in Designate controllers

omersch381:
So far, MariaDB account and secret were added to specs and types, and were used in designate_controller.go, but weren't used in the rest of the controllers.

This patch uses MariaDB in the rest of the controllers.


omersch381:
Is replaced by a few other PRs

PR #203, opened 2024-08-01
Add TLS field to Designate Operator API

omersch381:
This patch adds the TLS field to the Designate Operator API. The expectation is to set it correctly on Openstack Operator and then use it on Designate Operator.


api/v1beta1/designatebackendbind9_types.go
```diff
@@ -63,6 +64,10 @@ type DesignateBackendbind9SpecBase struct {
 	// +kubebuilder:validation:Optional
 	// ServiceAccount - service account name used internally to provide Designate services the default SA name
 	ServiceAccount string `json:"serviceAccount"`
+
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// TLS - Parameters related to the TLS
+	TLS tls.Ca `json:"tls,omitempty"`
```
beagles, line 70: Strictly speaking, I don't think we need this here but it might not hurt in the future. This would be a good thing to run by @johnsom

omersch381: I agree, in the following commit I removed the bind9 TLS parts, I will remove it here as well.

PR #211, opened 2024-08-13
Fix mdns replicas nil

omersch381:
Designate Mdns panicked after trying to access addresses in a few lines where it got nil pointer.

This patch fixes it by making sure the address it not nil.


omersch381:
Aborted by trigger plugin (?)

beagles (commented):
/lgtm

PR #214, opened 2024-08-19
Remove backendtype uses

beagles:
We do not support, nor do we have any plan to support DNS servers other than bind9 for the forseeable future. While we cannot remove it from the API, this patch removes it from the places it is used to avoid confusion or errorneous values from causing unexpected behaviour.


omersch381:
I also do not see any reason to have such config option if we have no plans to support other DNS backends.

PR #220, opened 2024-09-10
Generate an rndc key per bind instance

omersch381:
This PR adds the option of rndc key files generation. Given the bind replicas value, it allows Designate operator to either create or delete rndc/bind secrets.

### DIFF
--- a/controllers/designate_controller.go
+++ b/controllers/designate_controller.go
@@ -23,6 +23,7 @@ import (
 
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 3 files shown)


omersch381:
some of the functionality described in the ticket is still missing

controllers/designate_controller.go
```diff
@@ -1118,14 +1157,88 @@ func (r *DesignateReconciler) generateServiceConfigMaps(
...
 }
 
+func (r *DesignateReconciler) createAndStoreRndcKeySecrets(
+	ctx context.Context,
+	h *helper.Helper,
+	instance *designatev1beta1.Designate,
+	cmLabels map[string]string,
+	numOfSecretsToCreate int,
+) error {
+	for i := 0; i < numOfSecretsToCreate; i++ {
```
gthiemonge, line 1175: you always start at 0, so if 3 secrets exist (secret-0, secret-1, secret-2) and you need an additional one, it creates secret-0 intead of secret-3?

controllers/designate_controller.go
```diff
@@ -1059,9 +1063,44 @@ func (r *DesignateReconciler) generateServiceConfigMaps(
...
+	// Filter secrets with the name prefix "designate-bind-key"
+	filteredSecrets := []corev1.Secret{}
+	for _, secret := range secretList.Items {
+		if strings.HasPrefix(secret.Name, designate.DesignateBindKeySecretPrefix) {
+			filteredSecrets = append(filteredSecrets, secret)
+		}
+	}
+
+	numOfBindSecrets := len(filteredSecrets)
+	secretsDelta := int(math.Abs(float64(numOfBindSecrets) - float64(replicas)))
```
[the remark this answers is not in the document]
omersch381: To use the math.Abs function I need to use 2 floats. I can also multiply it by -1 if it is less than 0, but anyway, I am open to any other ideas.

controllers/designate_controller.go
omersch381: Yes, I missed that, thank you. I should use the replica count instead of the delta.

omersch381:
There is an error in the designate operator logs.

controllers/designate_controller.go
```diff
@@ -1059,9 +1062,59 @@ func (r *DesignateReconciler) generateServiceConfigMaps(
...
+			Data: keysMap,
+		}
+		if err := h.GetClient().Create(ctx, secret); err != nil {
+			if !k8s_errors.IsAlreadyExists(err) {
+				return err
+			}
+		}
+	}
+	numOfKeys := len(secret.Data)
+	keysDelta := int(math.Abs(float64(numOfKeys) - float64(replicas)))
```
gthiemonge, line 1100: now that keysDelta is only used in the numOfKeys > replicas case, you don't need all those conversions
you can directly pass (numOfKeys - replicas)

controllers/designate_controller.go
```diff
@@ -1059,9 +1059,55 @@ func (r *DesignateReconciler) generateServiceConfigMaps(
 	// - %-config configmap holding minimal designate config required to get the service up, user can add additional files to be added to the service
 	// - parameters which has passwords gets added from the ospSecret via the init container
 	//
+	Log := r.GetLogger(ctx)
 
 	cmLabels := labels.GetLabels(instance, labels.GetGroupLabel(designate.ServiceName), map[string]string{})
+	replicas := int(*instance.Spec.DesignateBackendbind9.Replicas)
+	secretName := designate.DesignateBindKeySecret
```
beagles, line 1066: nit: where. this is only used once you don't really need a temp var for it.

controllers/designate_controller.go
```diff
@@ -1059,9 +1059,55 @@ func (r *DesignateReconciler) generateServiceConfigMaps(
...
+	// Fetch the secret, create it if not found
+	_, err := controllerutil.CreateOrUpdate(ctx, h.GetClient(), secret, func() error {
+		secret.Labels = cmLabels
+		if secret.StringData == nil {
+			secret.StringData = make(map[string]string)
+		}
+
+		newKeysMap := make(map[string]string)
+
+		for i := 0; i < replicas; i++ {
```
beagles, line 1084: Can you check to make sure this works as expected with a scale out?

controllers/designate_controller.go
```diff
@@ -1059,9 +1059,55 @@ func (r *DesignateReconciler) generateServiceConfigMaps(
...
+			keyName := fmt.Sprintf("rndc-key-%v", i)
```
beagles, line 1085: Can you put the fmt string in a constant in the designate pkg or similar? It's kind of a shared detail.

beagles (changes requested):
Just a couple of minor nits/changes but looks good.

controllers/designate_controller.go
omersch381: ok, I will remove the secretName variable, thanks.

controllers/designate_controller.go
omersch381: Sure, sounds good to me, I will do that.

controllers/designate_controller.go
```diff
@@ -1059,9 +1059,54 @@ func (r *DesignateReconciler) generateServiceConfigMaps(
...
+
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      designate.DesignateBindKeySecret,
+			Namespace: instance.Namespace,
+		},
+	}
+
+	// Fetch the secret, create it if not found
+	_, err := controllerutil.CreateOrUpdate(ctx, h.GetClient(), secret, func() error {
```
gthiemonge, line 1075: Why don't you use EnsureSecret for creating it?
https://github.com/openstack-k8s-operators/lib-common/blob/814ac3823abe9baa929a74cc8e03086d650f5981/modules/common/secret/secret.go#L281-L286

https://github.com/search?q=repo%3Aopenstack-k8s-operators%2Fdesignate-operator%20EnsureSecrets&type=code

gthiemonge: Note that EnsureSecret aslo updates the hash that can be used to recreate the pods when the secret is updated.

beagles: It could be done that way. You first GetSecret.. if it doesn't exist, create the stringData map, create util.Template with the CustomData (I think) set to the that map and call EnsureSecrets with that.. If does exist, you do the work that's implemented in the mutator, getting the original contents of the secret first and creating/omitting entries as required and then you call EnsureSecrets. In some respects what Omer had done here is simpler/more straightforward. The only thing missing is to keep a handle of the secret and call secret.Hash() on it.

omersch381:
I tried reading from secret.StringData which is always nil, so I just changed it all to secret.Data which was easier for me.

PR #223, opened 2024-09-20
Add secret hash to worker hash

omersch381:
So far we had the rndc secret mounted on the worker pods, but any change to the existing secret would not modify the mounted rndc keys.
    
This patch adds the secret's hash to the worker's hash, which will "synchronize" the mounted keys according to the secret's state.

### DIFF
--- a/controllers/designateworker_controller.go
+++ b/controllers/designateworker_controller.go
@@ -487,7 +487,7 @@ func (r *DesignateWorkerReconciler) reconcileNormal(ctx context.Context, instanc
 	// create hash over all the different input resources to identify if any those changed
 	// and a restart/recreate is required.
 	//
-	inputHash, hashChanged, err := r.createHashOfInputHashes(ctx, instance, configMapVars)
+	inputHash, hashChanged, err := r.createHashOfInputHashes(ctx, helper, instance, configMapVars)
 	if err != nil {
 		instance.Status.Conditions.Set(condition.FalseCondition(
 			condition.ServiceConfigReadyCondition,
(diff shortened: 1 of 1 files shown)


controllers/designateworker_controller.go
```diff
@@ -487,6 +487,25 @@ func (r *DesignateWorkerReconciler) reconcileNormal(ctx context.Context, instanc
 	// create hash over all the different input resources to identify if any those changed
 	// and a restart/recreate is required.
 	//
+
+	// If DesignateBindKeySecret exists, add its hash to status hash
+	rndcSecret := &corev1.Secret{}
```
gthiemonge, line 492: I'm wondering if it can be integrated directly into createHashOfInputHashes (called line 509) which does almost the same thing.
I guess you only have to do
`configMapVars[designate.DesignateBindKeySecret] = env.SetValue(secretHash)`

omersch381: I have just tried it and it works. That method indeed does almost the same thing. Thanks

controllers/designateworker_controller.go
```diff
@@ -798,12 +798,28 @@ func (r *DesignateWorkerReconciler) generateServiceConfigMaps(
...
+		Name:      designate.DesignateBindKeySecret,
+		Namespace: instance.Namespace,
+	}, rndcSecret)
+	if err == nil {
+		secretHash, err := secret.Hash(rndcSecret)
+		if err != nil {
+			return secretHash, changed, err
+		}
+		envVars[designate.DesignateBindKeySecret] = env.SetValue(secretHash)
+	}
```
beagles, line 821: Shouldn't this return the err if it's not nil?

beagles (changes requested):
You'll also need to rebase this PR once 222 merges I think

controllers/designateworker_controller.go
omersch381: Yes it should, I will add it.

omersch381:
https://github.com/openstack-k8s-operators/designate-operator/pull/227 got merged and had this commit as well.

PR #224, opened 2024-09-20
Update bind configuration

beagles:
This patch gets bind up and running as a statefulset, tidying up the configuration files and making use of the per-bind rndc key generation implemented in an earlier patch.

Certain elements will require some further strict testing such as injecting name server options and setting up logging customizations.


templates/designatebackendbind9/bin/namedpoddetails.py
```diff
@@ -0,0 +1,38 @@
...
+
+import os
+import sys
+
+def getIPFromMap(name):
+    # TODO(beagles): we need predicatble IPs
+    return ''
+
+# Listen on all interfaces by default.
+ipVersion = os.environ.get('IPVERSION')
```
omersch381, line 25: os.environ.get is case sensitive, we save it as "IPVersion"

templates/designatebackendbind9/bin/namedpoddetails.py
```diff
@@ -0,0 +1,38 @@
...
+if ipVersion and ipVersion == 6:
+    podListenIP = '::'
+else:
+    podListenIP = '*'
+
+podname = os.environ.get('POD_INDEX').strip()
```
omersch381, line 31: same about case in os.environ.get

templates/designatebackendbind9/config-named/logging.conf
```diff
@@ -0,0 +1,29 @@
+logging {
+    channel default_channel {
+        file "/var/log/bind/esignate-bind.log";
```
omersch381, line 3: typo esignate-bind.log

templates/designatebackendbind9/config/named.conf
```diff
@@ -0,0 +1,5 @@
+include "/etc/named/rndc.key";
+include "/etc/named/rndc.conf";
+include "/etc/named/options.conf";
+include "/etc/named.rfc1912.zones";
+include "/etc/named/logging.conf";
```
omersch381, line 5: In the old templates/designate/config/named.conf we also had 
include "/etc/named.root.key";

omersch381 (commented):
A few comments and questions I had while going over the draft

templates/designatebackendbind9/bin/namedpoddetails.py
beagles: I don't think we store that at all yet (it would be done in the pod spec). You are thinking about the param that is used for generating the conf- not what is set in the environment. I should probably remove this script from the PR because it isn't used and is meaningless until we have a utility container for predictable IPs.

pkg/designatebackendbind9/volumes.go
```diff
@@ -0,0 +1,156 @@
...
+		},
+		{
+			Name:      namedConfigVolume,
+			MountPath: "/var/lib/config-data/default/named",
+			ReadOnly:  false,
+		},
+		{
+			Name:      mergedConfigVolume,
+			MountPath: "/var/lib/config-data/merged",
+			ReadOnly:  false,
```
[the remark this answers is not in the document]
beagles: That's because in line 112 is for the initcontainer and it populates the merged directory. The contents of this path when mounted into the service are meant to be readonly.

pkg/designatebackendbind9/volumes.go
```diff
@@ -0,0 +1,156 @@
...
+
+func getServicePodVolumeMounts(persistentData string) []corev1.VolumeMount {
+	return []corev1.VolumeMount{
+		{
+			Name:      mergedConfigVolume,
+			MountPath: "/var/lib/config-data/merged",
+			ReadOnly:  true,
+		},
+		{
+			Name:      mergedConfigVolume,
```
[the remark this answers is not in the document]
beagles: it is not. The second is a subpath mount and is used to place the kolla json file in the right place.

templates/designatebackendbind9/config/named.conf
beagles: I'll double check

templates/designatebackendbind9/config-named/logging.conf
beagles: nice catch! fixing.

templates/designatebackendbind9/bin/namedpoddetails.py
beagles: removing file as not relevant to this PR

pkg/designatebackendbind9/volumes.go
```diff
@@ -0,0 +1,156 @@
...
+package designatebackendbind9
+
+import (
+	designate "github.com/openstack-k8s-operators/designate-operator/pkg/designate"
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	scriptVolume       = "designatebackendbind9-scripts"
+	dataVolume         = "designatebackendbind9-bind-volume"
```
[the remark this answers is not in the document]
beagles: removing!

pkg/designatebackendbind9/service.go
```diff
@@ -7,25 +7,31 @@ import (
...
+	servicePorts := make([]corev1.ServicePort, 2)
+	servicePorts[0] = corev1.ServicePort{
+		Name:     "named",
+		Port:     53,
+		Protocol: "UDP",
+	}
+	servicePorts[1] = corev1.ServicePort{
+		Name:     "named",
+		Port:     53,
+		Protocol: "tcp",
```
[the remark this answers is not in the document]
beagles: agreed, fixing.

config/samples/designate_v1beta1_designate_tls.yaml
```diff
@@ -46,6 +48,8 @@ spec:
     customServiceConfig: |
       [DEFAULT]
       debug = true
+    storageRequest: 1G
+    storageClass: local-storage
```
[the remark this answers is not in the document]
beagles: ah fixing

templates/designatebackendbind9/config/named.conf
beagles: We do not have that in the tripleo config so I'm leaving out.

beagles:
held for a bit to finish a few extra tests.

templates/designatebackendbind9/config-named/options.conf
```diff
@@ -0,0 +1,38 @@
...
+
+        # The default for named is to listen on everything and it's controlled
+        # by different settings for different IP versions. If the objective is to
+        # restrict to listening on a given address, we need to explicitly turn
+        # listening *off* for the different IP version.
+
+        # TODO: The '*'s need to be replaced by actual addresses.
+{{ if eq .IPVersion "4" }}
+        listen-on port 53 { *; };
+        listen-on-v6 { none; };
```
omersch381, line 13: Are these mutually exclusive on OSP18? I am looking at my named.conf file on my devstack and I see
listen-on port 53 { 192.168.124.114; };
listen-on-v6 port 53 { ::1; };

omersch381: I also see this https://opendev.org/openstack/designate/src/branch/master/devstack/designate_plugins/backend-bind9#L167

templates/designatebackendbind9/config-named/rndc.conf
```diff
@@ -0,0 +1,7 @@
+// The current PODs bind9 address is set by the initcontainer.
+// WARNING: if the required information is not available
+// the bind service will be configured to listen to all IPv4
+// addresses
+controls {
+        inet * port 953 allow { {{.AllowCIDR}}; } keys { "rndc-key"; };
```
omersch381, line 6: Maybe we should have a comment like
TODO: The '*'s need to be replaced by actual addresses.
after inet
(I see only one address in my named.conf after inet)

templates/designatebackendbind9/config-named/options.conf
```diff
@@ -0,0 +1,38 @@
...
+             generated in place here. This is necessary as apparently you cannot have
+             mulitple 'options' sections in bind9 configuration */}}
+{{  if .CustomBindOptions }}
+    {{ range $index, $val := .CustomBindOptions }}
+            {{ $val }}
+    {{ end }}
+{{ end }}
+
+        minimal-responses yes;
+        multi-master yes;
```
omersch381, line 34: I wonder why would we want this one to be "yes"? I don't see it on the designate repo, and the default value is "no"
https://bind9.readthedocs.io/en/latest/reference.html#namedconf-statement-multi-master

omersch381 (commented):
Thanks Brent, a few questions

controllers/designatebackendbind9_controller.go
```diff
@@ -692,6 +638,70 @@ func (r *DesignateBackendbind9Reconciler) generateServiceConfigMaps(
...
+		templateParameters["IPVersion"] = "4"
+	} else {
+		templateParameters["IPVersion"] = "6"
+	}
+	templateParameters["AllowCIDR"] = cidr
+	// This will need to be replaced by custom config for named.
+	templateParameters["EnableQueryLogging"] = false
+	templateParameters["CustomBindOptions"] = instance.Spec.CustomBindOptions
+
+	// TODO: we need the rndc key value and pod addr but those are going to be supplied by the init container and
```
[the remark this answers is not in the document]
beagles: The rndc key is already mounted. The pod addr can be mounted once it's merged in.

pkg/designatebackendbind9/deployment.go
```diff
@@ -123,28 +99,16 @@ func Deployment(
...
-							Image: instance.Spec.ContainerImage,
-							SecurityContext: &corev1.SecurityContext{
-								RunAsUser: &rootUser,
-							},
-							Env:           env.MergeEnvs([]corev1.EnvVar{}, envVars),
-							VolumeMounts:  volumeMounts,
-							Resources:     instance.Spec.Resources,
-							StartupProbe:  startupProbe,
-							LivenessProbe: livenessProbe,
+							Name:           instance.Name,
```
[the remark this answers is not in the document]
beagles: good eyes, I'll change that back to be consistent with the others.

pkg/designatebackendbind9/volumes.go
```diff
@@ -0,0 +1,155 @@
...
+			ReadOnly:  false,
+		},
+		{
+			Name:      scriptVolume,
+			MountPath: "/usr/local/bin/container-scripts",
+			ReadOnly:  true,
+		},
+		{
+			Name:      rndcKeys,
+			MountPath: "/var/lib/config-data/keys",
```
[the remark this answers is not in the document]
beagles: Not necessary. IIRC you are mounting them right where they are going to be used. I'm mounting them in a place out of the way so it can be picked up and copied into the necessary location. The bind only uses one of them, where the workers uses all of them.

templates/designatebackendbind9/bin/init.sh
```diff
@@ -0,0 +1,52 @@
...
+    rndc_prefix="${RNDC_PREFIX}-"
+fi
+
+# get the index off of the pod name
+set -f
+name_parts=(${POD_NAME//-/ })
+pod_index="${name_parts[-1]}"
+rndc_key_filename="/var/lib/config-data/keys/${rndc_prefix}${pod_index}"
+if [[ -f "${rndc_key_filename}" ]]; then
+    cp ${rndc_key_filename} /var/lib/config-data/merged/named/rndc.key
```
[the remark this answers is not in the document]
beagles: Nope see above. For bind it only uses one of them and really you don't want all of them in the bind container.

templates/designatebackendbind9/config-named/options.conf
beagles: We need the predictable IPs to set these to what they are going to be.

templates/designatebackendbind9/config-named/options.conf
beagles: Hrm. We'll need to check the tripleo install and the jira issue. That's the reference we should use.

beagles: Yeah, multi-master is yes in tripleo. IIRC the "master's" in this case are the mdns servers, which there are many.

templates/designatebackendbind9/config-named/options.conf
beagles: Sorry, I misunderstood the question. Yes, there are mutually exclusive. I don't think dualstack support is on the road map for now.

PR #227, opened 2024-09-27
Implement predictable IP generation for miniDNS

omersch381:
The designate miniDNS needs a predictable IP mechanism, as it is a
daemonset and will use the list of available workers to generate a list
of IPs that are valid for the designate network attachment's CIDR range
but do not overlap with the address range defined in the attachment's
IPAM config.

This patch adds it.


beagles:
This looks fine. We won't be able to properly test until we get our utility container image in place.

PR #229, opened 2024-09-30
Generate predictable IPs for bind stateful set

omersch381:
Generate and manage predictable IPs for each member of the bind
statefulset and store in it's own predictable IPs config map.

The reason it needs to be it's own map is that we do not want to cause
minidns to restart if a scaling change occurs on mind - nor do we want
minidns scale changes to affect bind pods.

### DIFF
--- a/controllers/designate_controller.go
+++ b/controllers/designate_controller.go
@@ -714,6 +714,7 @@ func (r *DesignateReconciler) reconcileNormal(ctx context.Context, instance *des
 	}
 	Log.Info("Deployment API task reconciled")
 
+	// Handle Mdns predictable IPs configmap
 	nad, err := nad.GetNADWithName(ctx, helper, instance.Spec.DesignateNetworkAttachment, instance.Namespace)
 	if err != nil {
 		return ctrl.Result{}, err
@@ -722,31 +723,6 @@ func (r *DesignateReconciler) reconcileNormal(ctx context.Context, instance *des
 	networkParameters, err := designate.GetNetworkParametersFromNAD(nad)
 	if err != nil {
 		return ctrl.Result{}, err
-	}
-
-	nodeConfigMap := &corev1.ConfigMap{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      designate.MdnsPredIPConfigMap,
-			Namespace: instance.GetNamespace(),
-			Labels:    labels.GetLabels(instance, labels.GetGroupLabel(instance.ObjectMeta.Name), map[string]string{}),
-		},
-		Data: make(map[string]string),
-	}
-
-	// Look for existing config map and if exists, read existing data and match
-	// against nodes.
-	foundMap := &corev1.ConfigMap{}
-	err = helper.GetClient().Get(ctx, types.NamespacedName{Name: designate.MdnsPredIPConfigMap, Namespace: instance.GetNamespace()},
-		foundMap)
-	if err != nil {
-		if k8s_errors.IsNotFound(err) {
-			Log.Info(fmt.Sprintf("Ip map %s doesn't exist, creating.", designate.MdnsPredIPConfigMap))
-		} else {
-			return ctrl.Result{}, err
-		}
-	} else {
-		Log.Info("Retrieved existing map, updating..")
-		nodeConfigMap.Data = foundMap.Data
 	}
 
 	//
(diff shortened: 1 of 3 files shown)


omersch381:
TODO: a few comments from beagles: 
1. predictable IPs for bind’s cannot be tied to the node list, they are tied to replica count and should not reference nodes at all.
2. you need to avoid IP conflicts either by examining both config maps and extracting the used addresses before you start allocating IPs -OR- use two non-overlapping ranges, one for each  case. I’m not sure what is easier. You still of course have to query the existing IPs, but it would be per-range

controllers/designate_controller.go
```diff
@@ -713,6 +714,99 @@ func (r *DesignateReconciler) reconcileNormal(ctx context.Context, instance *des
...
+		return ctrl.Result{}, err
+	}
+
+	var nodeNames []string
+	for _, node := range nodes.Items {
+		nodeNames = append(nodeNames, fmt.Sprintf("mdns_%s", node.Name))
+	}
+
+	allocatedIPs := make(map[string]bool)
+	updatedMap, allocatedIPs, err := r.allocatePredictableIPs(ctx, predictableIPParams, nodeNames, nodeConfigMap.Data, allocatedIPs)
```
beagles, line 765: IIUC this will work the first time around, but if the number of nodes for minidns nodes runs, you are going to run into collisions with the IPs allocated for the bind hosts. What you need to do is get ALL of the allocated IPs for both config maps first.

beagles (changes requested):

controllers/designate_controller.go
omersch381: Thanks, I have just added fetching both configmaps before I allocate predictable IPs.

controllers/designate_controller.go
```diff
@@ -1082,6 +1077,73 @@ func (r *DesignateReconciler) reconcileNormal(ctx context.Context, instance *des
...
+		}
+	} else {
+		Log.Info("Retrieved existing map, updating..")
+		nodeConfigMap.Data = foundMap.Data
+	}
+
+	return nodeConfigMap, nil
+}
+
+func (r *DesignateReconciler) allocatePredictableIPs(ctx context.Context, predictableIPParams *designate.NADIpam, nodeNames []string, existingMap map[string]string, allocatedIPs map[string]bool) (map[string]string, map[string]bool, error) {
```
beagles, line 1110: nit: do you mind changing the arg name "nodeNames" to something else as it doesn't really match anymore?

controllers/designate_controller.go
```diff
@@ -1082,6 +1077,73 @@ func (r *DesignateReconciler) reconcileNormal(ctx context.Context, instance *des
...
+
+	updatedMap := make(map[string]string)
+	var predictableIPsRequired []string
+
+	// First scan existing allocations so we can keep existing allocations.
+	// Keeping track of what's required and what already exists. If a node is
+	// removed from the cluster, it's IPs will not be added to the allocated
+	// list and are effectively recycled.
+	for _, nodeName := range nodeNames {
+		if ipValue, ok := existingMap[nodeName]; ok {
```
beagles, line 1121: IIUC, this block isn't really needed anymore. If we do remove it, it would be nice to move the logging up to where it is originally initialized.

omersch381: we still need to fill the updatedMap, afterwards we return it and set 
mdnsConfigMap.Data = updatedMap
bindConfigMap.Data = updatedBindMap

but I don't think we need line 1123 anymore, I will remove it.

controllers/designate_controller.go
omersch381: When we call that function for mdns, we do get the node list and pass their names. I can change it to ip_holders, maybe?
because afterwards we get something like this:
~ > oc get configmaps designate-mdns-ip-map -o json | jq -r '.data'; oc get configmaps designate-bind-ip-map -o json | jq -r '.data'
{
  "mdns_crc-vlf7c-master-0": "172.28.0.97"
}
{
  "bind_address_0": "172.28.0.98",
  "bind_address_1": "172.28.0.99"
}

controllers/designate_controller.go
beagles: I think you're right here actually - your initializing the updated map here, not the existing map.

controllers/designate_controller.go
beagles: Right, it's just a nit that that in the bind case it isn't node names anymore.

beagles (approved):
/lgtm

PR #235, opened 2024-10-15
Add pools.yaml generator

omersch381:
This commit adds a module that will generate an "up-to-date" pools.yaml file. It will use both Bind9, Mdns and NsRecords configmaps as input to generate that file.


omersch381:
This PR misses (at least) 2 things:

1. Kuttl tests
2. Calling the Hash function for each configmap that might have changed, so an up-to-date pools.yaml file will be generated

omersch381:
error: build error: Failed to push image: trying to reuse blob sha256:4d5d1cbd7ece41ce278c26338e01e2b82e1861b820ca052da9f3e0b16815358f at destination: pinging container registry quay.rdoproject.org: Get "https://quay.rdoproject.org/v2/": dial tcp 38.129.56.158:443: i/o timeout

pkg/designate/generate_bind9_pools_yaml.go
```diff
@@ -0,0 +1,215 @@
...
+				Port: 53,
+			})
+
+			target := Target{
+				Type:        "bind9",
+				Description: fmt.Sprintf("BIND9 Server %d (%s)", rndcKeyNum, bindIP),
+				Masters: []Master{
+					{
+						Host: mdnsIP,
+						Port: 16000, // TODO oschwart: change it to the real value
```
beagles, line 115: just change to 53 I think

pkg/designate/generate_bind9_pools_yaml.go
```diff
@@ -0,0 +1,215 @@
...
+	RNDCPort       int
+	RNDCConfigFile string
+}
+
+type CatalogZone struct {
+	FQDN    string
+	Refresh string
+}
+
+func GeneratePoolsYamlFile(pooslData []PoolData) (string, error) {
```
beagles, line 75: probably meant poolsData

pkg/designate/generate_bind9_pools_yaml.go
```diff
@@ -0,0 +1,215 @@
...
+		err = tmpl.Execute(&buf, pool)
+		if err != nil {
+			return "", err
+		}
+	}
+
+	return buf.String(), nil
+}
+
+const poolsTemplate = `
```
beagles, line 170: this probably should not be in code but in a file in the templates directory. It's debatable whether it is any benefit outside of development (able to tweak template without recompiling) or checking a change at runtime but it will be consistent with other things that do this.

pkg/designate/generate_bind9_pools_yaml.go
omersch381: Yes, I don't return a file, thanks.
Maybe poolsYamlContent? I return a string that could be saved as the pools.yaml file.

pkg/designate/generate_bind9_pools_yaml.go
omersch381: Sounds good to me, I will move poolsTemplate to templates directory and import it from there.

pkg/designate/generate_bind9_pools_yaml.go
omersch381: on my devstack environment it was 5354, and the values that I have here I copied from a 17.1 host. I am not really sure what value should be here tbh.

pkg/designate/generate_bind9_pools_yaml.go
omersch381: I will change it to poolsYamlData

pkg/designate/generate_bind9_pools_yaml.go
```diff
@@ -0,0 +1,170 @@
...
+		nsRecords := []NSRecord{}
+		for hostname, priority := range poolData.NsRecordsMap {
+			nsRecord := NSRecord{
+				Hostname: hostname,
+				Priority: priority,
+			}
+			nsRecords = append(nsRecords, nsRecord)
+		}
+
+		// As far as I understand, we will only have one mdnsIP (primary DNS server)
```
beagles, line 101: these aren't primaries, but the predictable IPs for the mDns servers (which are masters)

pkg/designate/generate_bind9_pools_yaml.go
```diff
@@ -0,0 +1,170 @@
...
+		for _, bindIP := range poolData.BindMap {
+			nameservers = append(nameservers, Nameserver{
+				Host: bindIP,
+				Port: 53,
+			})
+
+			target := Target{
+				Type:        "bind9",
+				Description: fmt.Sprintf("BIND9 Server %d (%s)", rndcKeyNum, bindIP),
+				Masters: []Master{
```
beagles, line 122: this should be a master per mDNS instance. For example our default production configuration would probably be 3 or more.

tests/kuttl/tests/basic/ns-records-configmap/02-assert.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: designate-ns-records-params
+data:
+  ns_records: |
```
beagles, line 6: I'm not sure you need the pipe here.

pkg/designate/generate_bind9_pools_yaml.go
omersch381: I changed it, I will upload another version.

pkg/designate/generate_bind9_pools_yaml.go
omersch381: I changed it, I will upload another version.

tests/kuttl/tests/basic/ns-records-configmap/02-assert.yaml
omersch381: I tried it without the pipe and it fails, so I will leave it there.

controllers/designate_controller.go
```diff
@@ -1339,6 +1389,127 @@ func (r *DesignateReconciler) generateServiceConfigMaps(
 	return nil
 }
 
+func (r *DesignateReconciler) getEnvVars(
```
beagles, line 1392: I'm not entirely sure what we are doing here... I think we might be getting tangled between the old way of setting a pile of configuration via a bash script and the other through templates..

omersch381: Yes, I totally agree. So after Brent and I went over the code, we decided we should have an additional hash, PoolsYamlHash, that won't be depended on the environment variables. I have just uploaded a new version.

beagles (commented):
/lgtm

PR #237, opened 2024-10-17
fix allWatchFields for the different controllers

stuggi:


stuggi:
error: build error: Failed to push image: trying to reuse blob sha256:4d5d1cbd7ece41ce278c26338e01e2b82e1861b820ca052da9f3e0b16815358f at destination: pinging container registry quay.rdoproject.org: Get "https://quay.rdoproject.org/v2/": dial tcp 38.129.56.158:443: i/o timeout

PR #247, opened 2024-11-05
Add pool update job

omersch381:


controllers/designate_controller.go
```diff
@@ -831,7 +831,7 @@ func (r *DesignateReconciler) reconcileNormal(ctx context.Context, instance *des
 		}
 		Log.Info(fmt.Sprintf("pools.yaml content is\n%v", poolsYaml))
 		updatedPoolsYaml := make(map[string]string)
-		updatedPoolsYaml[designate.PoolsYamlsConfigMap] = poolsYaml
+		updatedPoolsYaml["pools-yaml-content"] = poolsYaml
```
beagles, line 834: I'm curious why the switch to from using a constant to a hard coded string here?

pkg/designate/pool_update.go
```diff
@@ -0,0 +1,132 @@
...
+				},
+			},
+		},
+	})
+
+	volumes = append(volumes, corev1.Volume{
+		Name: DesignateConfigVolume,
+		VolumeSource: corev1.VolumeSource{
+			Secret: &corev1.SecretVolumeSource{
+				SecretName: "designate-api-config-data",
```
beagles, line 50: I wonder if this is the right secret name. The pool update job probably has more in common with central. I suggest we retest with the merged redis patch to make sure that this still works.

pkg/designate/pool_update.go
```diff
@@ -0,0 +1,132 @@
...
+				Items: []corev1.KeyToPath{
+					{
+						Key:  "designate.conf",
```
beagles, line 53: Is this part actually needed. IIUC the file designate.conf will be there already.

pkg/designate/pool_update.go
```diff
@@ -0,0 +1,132 @@
...
+				},
+			},
+		},
+	})
+
+	volumes = append(volumes, corev1.Volume{
+		Name: "rabbitmq-certs",
+		VolumeSource: corev1.VolumeSource{
+			Secret: &corev1.SecretVolumeSource{
+				SecretName: "cert-rabbitmq-svc",
```
beagles, line 65: I wonder if this is correct to assume this name.

pkg/designate/pool_update.go
omersch381: Sounds good to me. I think that I implemented it before I had access to the redis PR. So now that we had an "upgraded" version of central, let me change it and test.

pkg/designate/pool_update.go
omersch381: Now I guess that it will, before I execed the pod and didn't find the correct one. So if I have the correct designate.conf there, I guess I could remove that volume mounting.

pkg/designate/pool_update.go
omersch381: Done

pkg/designate/pool_update.go
beagles: Actually I think I was wrong here. IIUC the jobspec doesn't have an init container which is part of what does all of the config stuff, so I think this correct. One thing to keep in mind -- and I suspect it isn't an issue because central is needed for this to run -- is this secret isn't available until after the code that generates it is run in the designatecentral_controller.

pkg/designatecentral/deployment.go
```diff
@@ -43,8 +43,9 @@ func Deployment(
 	// designateUser := int64(42411)
 	// designateGroup := int64(42411)
 
-	volumes := designate.GetVolumes("designate-central")
-	volumeMounts := designate.GetVolumeMounts("designate-central")
+	serviceName := fmt.Sprintf("%s-central", designate.ServiceName)
```
beagles, line 46: Does designate central actually use the pools.yaml file?

omersch381: Hmm I don't think so, probably only the job should use it. I will comment it from the central pod related stuff

controllers/designate_controller.go
```diff
@@ -864,9 +864,36 @@ func (r *DesignateReconciler) reconcileNormal(ctx context.Context, instance *des
...
+			_, err = poolUpdatejob.DoJob(ctx, helper)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+			instance.Status.Hash[designatev1beta1.PoolUpdateHash] = poolUpdatejob.GetHash()
+			err = r.Client.Status().Update(ctx, instance)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+			Log.Info("Pool update job completed successfully")
```
beagles, line 896: Looking at where this happens in the reconcile order: IIRC the pool update *requires* that central be running before it does. So it might make more sense to a.) move this to the bottom of the reconcile function and b.) check central's status before running.

beagles:
I think the functional tests can't really test the pools thing unless you can fake out the replica count for designate central

PR #253, opened 2024-11-22
Expose bind9 endpoints

omersch381:
This patch creates series of services and map them to the statefulset members using the podname.

Each service has an IP from a metallb pool.


omersch381:
note: it looks like the services are created with a metalLB ip, but I am not sure if we want an endpoint as well
~ > oc get svc | grep designate                                                                 
designate-backendbind9-designate-backendbind9-0   LoadBalancer   10.217.4.244   172.17.0.81      53:32663/UDP,53:32663/TCP                        33m
designate-backendbind9-designate-backendbind9-1   LoadBalancer   10.217.5.180   172.17.0.82      53:30551/UDP,53:30551/TCP                        33m
designate-internal                                LoadBalancer   10.217.4.178   172.17.0.80      9001:31218/TCP                                   33m
designate-public                                  ClusterIP      10.217.4.100   <none>           9001/TCP                                         33m
designate-unbound                                 ClusterIP      10.217.4.212   <none>           53/TCP                                           33m

~ > oc get endpoints | grep designate                                                           
designate-backendbind9-0                  <none>                                                                    33m
designate-backendbind9-1                  <none>                                                                    33m
designate-internal                                10.217.0.201:9001                                                   33m
designate-public                                  10.217.0.201:9001                                                   33m
designate-unbound                                 10.217.0.194:53                                                     33m

pkg/designatebackendbind9/service.go
```diff
@@ -1,40 +1,66 @@
...
-		Selector: map[string]string{
-			"app": "bind",
+
+	return &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      podName,
+			Namespace: m.Namespace,
+			Labels:    labels,
+			Annotations: map[string]string{
+				"metallb.universe.tf/address-pool": "internalapi",
```
beagles, line 40: The address pool will need to be configurable.

pkg/designatebackendbind9/service.go
```diff
@@ -1,40 +1,66 @@
...
+const (
+	Bind9AppName = "designatebackendbind9"
+)
+
+// CreateBindServices creates a service for each bind9 pod
+func CreateBindServices(m *designatev1beta1.DesignateBackendbind9, replicas int32) []*corev1.Service {
+	services := make([]*corev1.Service, replicas)
+
+	for i := int32(0); i < replicas; i++ {
+		podName := fmt.Sprintf("%s-%d", m.Name, i)
```
beagles, line 20: This might not be correct for a mapping. We'll need to figure that out.

beagles (changes requested):
I think there's a couple of things we need to sort out with names, but this looks like an excellent start.

pkg/designatebackendbind9/service.go
omersch381: Done, I will upload another version.

api/v1beta1/designatebackendbind9_types.go
```diff
@@ -78,6 +78,11 @@ type DesignateBackendbind9SpecBase struct {
 	// +kubebuilder:validation:Optional
 	// StorageRequest
 	StorageRequest string `json:"storageRequest"`
+
+	// +kubebuilder:default="internalapi"
```
beagles, line 82: In retrospect, I don't think this is a good default. It may not be possible to to have a default really. What are your thoughts on leaving it empty  and only creating the services if a pool is provided?

omersch381: sounds good to me.

pkg/designatebackendbind9/service.go
beagles: let's leave as is - atm the pod name will be taken from the instance name. If that changes for some reason we can adapt it later.

omersch381:
PR is not relevant anymore.

PR #257, opened 2024-12-02
Change Unbound deployment type to StatefulSet

omersch381:
Change Unbound deployment type to StatefulSet


omersch381:
error: build error: Failed to push image: trying to reuse blob sha256:29291e31a76a7e560b9b7ad3cada56e8c18d50a96cca8a2573e4f4689d7aca77 at destination: pinging container registry quay.rdoproject.org: Get "https://quay.rdoproject.org/v2/": dial tcp 38.129.56.158:443: i/o timeout

stuggi:
do we expect that designate is already used and also the cleanup of the current deployments need to be handled?

omersch381:
Designate is still disabled by default in all of the CRs. It shouldn't be in use by anyone (yet).

PR #262, opened 2024-12-16
Add functional tests

omersch381:


api/v1beta1/conditions.go
```diff
@@ -21,8 +21,8 @@ import (
 
 // Designate Condition Types used by API objects.
 const (
-	// DesignateRabbitMqTransportURLReadyCondition Status=True condition which indicates if the RabbitMQ TransportURLUrl is ready
-	DesignateRabbitMqTransportURLReadyCondition condition.Type = "DesignateRabbitMqTransportURLReady"
+	// RabbitMqTransportURLReadyCondition Status=True condition which indicates if the RabbitMQ TransportURLUrl is ready
+	RabbitMqTransportURLReadyCondition condition.Type = "RabbitMqTransportURLReady"
```
beagles, line 25: For future reference:a PR that adds tests probably should avoid changing other code. I think it's a decent change in any case.

api/v1beta1/conditions.go
```diff
@@ -52,19 +52,19 @@ const ()
 // Common Messages used by API objects.
 const (
 	//
-	// DesignateRabbitMqTransportURLReady condition messages
+	//  RabbitMqTransportURLReady condition messages
 	//
-	// DesignateRabbitMqTransportURLReadyInitMessage
-	DesignateRabbitMqTransportURLReadyInitMessage = "DesignateRabbitMqTransportURL not started"
+	// RabbitMqTransportURLReadyInitMessage
+	RabbitMqTransportURLReadyInitMessage = "RabbitMqTransportURL not started"
```
beagles, line 58: Since you're changing them anyways, can you check the messages this against messages in other operators like nova? These were copy/pasted early on and the text messages aren't super helpful as they refer to constant names.

beagles (commented):
Good work! Just a question about the message text.

api/v1beta1/conditions.go
beagles: Yeah you are right. In fact it seems we should not be redefining these and instead use the conditions and messages from lib-common

omersch381:
hmm the only change I did was replacing designatev1beta.RabbitMqTransportURLReadyCondition with condition.RabbitMqTransportURLReadyCondition and it is not even used in the test.

PR #268, opened 2025-01-14
Add support for configuring unbound forwarders

omersch381:
This patch adds support for unbound forwarders configuration. The end user / operator should create a configmap similar to the one on demo/examples and the forwarder files will be mounted in the unbound pod.

### DIFF
--- a/demo/examples/unbound_forwarder_example.yaml
+++ b/demo/examples/unbound_forwarder_example.yaml
@@ -1,0 +1,15 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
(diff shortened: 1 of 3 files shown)


beagles:
Considering this is basically just adding an addtional config file via config map, DefaultConfigOverwrite should probably work for this. Maybe try that instead of adding special case for that.

beagles:
We are actually pretty much blocked by https://github.com/openstack-k8s-operators/openstack-operator/pull/1281

beagles:
Closing this pull request as the DefaultConfigOverwrite works great.

PR #275, opened 2025-01-24
Expose unbound endpoints

omersch381:
This patch creates series of services and map them to the statefulset members using the podname.

Each service has an IP from a metallb pool.


omersch381:
PR is not relevant anymore.

PR #277, opened 2025-01-28
Fix pool update job

omersch381:
The pool update job was not launched correctly because all of the job instances had the same name. As Kubernetes requires names be unique across all API versions of the same resource, the job was only triggered once.

This patch adds the jobs a unix timestamp suffix to overcome that problem.

It also sorts the pool resources to create a comparable hash between jobs.

### DIFF
--- a/controllers/designate_controller.go
+++ b/controllers/designate_controller.go
@@ -833,7 +833,6 @@ func (r *DesignateReconciler) reconcileNormal(ctx context.Context, instance *des
 	}
 
 	Log.Info("Bind configmap was created successfully")
-
 	if len(nsRecordsConfigMap.Data) > 0 && instance.Status.DesignateCentralReadyCount > 0 {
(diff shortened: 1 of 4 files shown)


controllers/designate_controller.go
```diff
@@ -831,7 +831,10 @@ func (r *DesignateReconciler) reconcileNormal(ctx context.Context, instance *des
 
 	Log.Info("Bind configmap was created successfully")
 
-	if len(nsRecordsConfigMap.Data) > 0 && instance.Status.DesignateCentralReadyCount > 0 {
+	bindReady := designatev1beta1.DesignateBackendbind9ReadyCondition == "DesignateBackendbind9Ready"
```
beagles, line 834: These aren't needed. You only really need to make sure central has something up.

omersch381: Right, I have just tested it, thanks. I will remove those.

PR #281, opened 2025-02-07
WIP Change nsRecords to API field instead of configmap

omersch381:


config/samples/ns_records_CR_example.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: core.openstack.org/v1beta1
+kind: OpenStackControlPlane
```
beagles, line 2: I think this kind of sample would only work if Kind: designate. Maybe consider putting this in the designate_v1beta1_designate.yaml file instead

omersch381: Done, thanks

PR #285, opened 2025-02-14
Add NsRecords API fields under designateAPI

omersch381:
So far we were using ns_records configmap to configure the pools ns records.

This patch adds the NsRecords API field as in the near future we plan to switch to those instead of configmaps.


beagles (commented):
/lgtm

PR #292, opened 2025-02-25
Config cleanup and pools yaml ordering

beagles:
These are some remaining cleanups and corrections as well as a fix to a problem that was discovered late in ad-hoc testing.

### DIFF
--- a/templates/designateunbound/config/unbound.conf
+++ b/templates/designateunbound/config/unbound.conf
@@ -1,11 +1,7 @@
 server:
 	do-daemonize: no
-	interface: {{ .ListenIP }}
-	access-control: {{ .ExternalNetCidr }} allow
-
-# XXX(beagles) figure out how to insert acl for the user's config. We can't
-# merge it here because it will just obliterate what's already there.
-
+        interface: 0.0.0.0
+        interface: ::0
 	log-queries: no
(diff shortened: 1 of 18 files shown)


templates/designateunbound/config/unbound.conf
```diff
@@ -1,11 +1,6 @@
 server:
 	do-daemonize: no
-	interface: {{ .ListenIP }}
-	access-control: {{ .ExternalNetCidr }} allow
-
-# XXX(beagles) figure out how to insert acl for the user's config. We can't
-# merge it here because it will just obliterate what's already there.
-
+        interface: 0.0.0.0
```
gthiemonge, line 3: is it supposed to be ipv6 compatible?

it reminds me something we had to do in octavia to support ipv4/ivp6: https://github.com/openstack-k8s-operators/octavia-operator/blob/ef470477c47639ca2b42f45723ae7f9810612bc3/templates/octaviaamphoracontroller/config/octavia.conf#L16
(note: it may not work in this config file)

beagles: Yeah I think I can add the IPv6 interface as well as IPv4... PODs seem to have some kind of IPv6 addr.

PR #305, opened 2025-03-14
Add support to NsRecords via CR

omersch381:
So far we only allowed using NsRecords via a ConfigMap.

While using nsRecordsConfigMap works, creating/modifying it does not trigger another run of the reconciliation loop. That means the user can modify the ns records config map and wait for a long time without the changes to take affect from Designate perspective.

This patch adds support for adding nsRecords via Designate CR/csv, which will take precedence over the ns records config map. If nsRecords is not configured in the Designate CR, it will fall back to the configmap.

### DIFF
--- a/controllers/designate_controller.go
+++ b/controllers/designate_controller.go
@@ -22,6 +22,7 @@ import (
 	"sort"
 	"time"
 
+	"gopkg.in/yaml.v2"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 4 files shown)


controllers/designate_controller.go
```diff
@@ -1158,6 +1159,32 @@ func (r *DesignateReconciler) reconcileNormal(ctx context.Context, instance *des
 	return ctrl.Result{}, nil
 }
 
+func (r *DesignateReconciler) getNSRecordsData(ctx context.Context, helper *helper.Helper, instance *designatev1beta1.Designate, labels map[string]string) (map[string]string, error) {
+	Log := r.GetLogger(ctx)
+
+	// First check if NS records are defined in the CR
+	if len(instance.Spec.NSRecords) > 0 {
+		Log.Info("Using NS records from CR", "count", len(instance.Spec.NSRecords))
```
beagles, line 1167: Would it be a bit tidier to redo the signature of GeneratePoolsYamlDataAndHash to take a slice of NSRecord, avoiding marshalling to yaml just to Unmarshal later? For doing the configmap case, you could marshal it into a slice of NSRecord? It saves as step and makes the preferred path simpler than the unpreferred.

omersch381:
functional tests passed on my host

PR #310, opened 2025-03-27
[OSPRH-14758] Fix controlNetworkName requires default value

omersch381:
So far we did not have a default value for controlNetworkName, which resulted in Designate network attachment not being found in some deployments.

This patch adds a hard-coded default value for Designate network attachment in case one was not defined in the CR.

### DIFF
--- a/controllers/designate_controller.go
+++ b/controllers/designate_controller.go
@@ -1665,6 +1665,12 @@ func (r *DesignateReconciler) mdnsStatefulSetCreateOrUpdate(ctx context.Context,
 		statefulSet.Spec.NodeSelector = instance.Spec.DesignateMdns.NodeSelector
 		statefulSet.Spec.TopologyRef = instance.Spec.DesignateMdns.TopologyRef
 
+		networkAttachment := "designate"
+		if instance.Spec.DesignateNetworkAttachment != "" {
+			networkAttachment = instance.Spec.DesignateNetworkAttachment
+		}
+		statefulSet.Spec.ControlNetworkName = networkAttachment
+
 		err := controllerutil.SetControllerReference(instance, statefulSet, r.Scheme)
 		if err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


controllers/designatebackendbind9_controller.go
```diff
@@ -788,6 +788,12 @@ func (r *DesignateBackendbind9Reconciler) generateServiceConfigMaps(
 	}
 
 	var nadInfo *designate.NADConfig
+	controlNetworkName := instance.Spec.ControlNetworkName
+
+	if controlNetworkName == "" {
```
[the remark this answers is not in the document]
omersch381: thanks! Fixed.

controllers/designate_controller.go
```diff
@@ -1664,6 +1664,7 @@ func (r *DesignateReconciler) mdnsStatefulSetCreateOrUpdate(ctx context.Context,
 		statefulSet.Spec.TLS = instance.Spec.DesignateAPI.TLS.Ca
 		statefulSet.Spec.NodeSelector = instance.Spec.DesignateMdns.NodeSelector
 		statefulSet.Spec.TopologyRef = instance.Spec.DesignateMdns.TopologyRef
+		statefulSet.Spec.ControlNetworkName = instance.Spec.DesignateNetworkAttachment
```
beagles, line 1667: You should check if DesignateNetworkAttachment and use  "designate" as the de-facto default if it is also empty.

omersch381: ok, done.

controllers/designate_controller.go
```diff
@@ -1665,6 +1665,12 @@ func (r *DesignateReconciler) mdnsStatefulSetCreateOrUpdate(ctx context.Context,
 		statefulSet.Spec.NodeSelector = instance.Spec.DesignateMdns.NodeSelector
 		statefulSet.Spec.TopologyRef = instance.Spec.DesignateMdns.TopologyRef
 
+		networkAttachment := "designate"
```
beagles, line 1668: it occurs to me we should've created a constant somewhere but this is fine.

PR #334, opened 2025-06-05
Migrate Designate Operator to uwsgi

omersch381:
mod_wsgi is being removed from upstream sources but OpenStack APIs are still using it downstream, which is very concerning.

This patch aims to replace mod_wsgi with uwsgi.

Depends-On: https://github.com/openstack-k8s-operators/tcib/pull/282

### DIFF
--- a/templates/designateapi/config/httpd.conf
+++ b/templates/designateapi/config/httpd.conf
@@ -46,11 +46,8 @@ TimeOut {{ $.TimeOut }}
     SSLCertificateKeyFile   "{{ $vhost.SSLCertificateKeyFile }}"
     {{- end }}
 
-    ## WSGI configuration
-    WSGIProcessGroup {{ $endpt }}
-    WSGIApplicationGroup %{GLOBAL}
-    WSGIPassAuthorization On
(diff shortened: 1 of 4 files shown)


templates/designateapi/config/httpd.conf
```diff
@@ -46,11 +46,8 @@ TimeOut {{ $.TimeOut }}
...
     {{- end }}
 
-    ## WSGI configuration
-    WSGIProcessGroup {{ $endpt }}
-    WSGIApplicationGroup %{GLOBAL}
-    WSGIPassAuthorization On
-    WSGIDaemonProcess {{ $endpt }} processes=5 threads=1 user=designate group=designate display-name={{ $endpt }}
-    WSGIScriptAlias / "/usr/bin/designate-api-wsgi"
+    # ProxyPass configuration for uwsgi
+    ProxyPass "/dns" "http://127.0.0.1:60053" retry=0
```
johnsom, line 50: I would recommend adding "acquire=1" here.

johnsom (changes requested):

omersch381:
Thank you guys for the comments. I will apply those if we decide eventually we continue with this direction (UWSGI).

PR #337, opened 2025-06-20
Sets the listen address to the predictable IP. 

beagles:
Have mDNS listen and use the predictable IP for apps expecting replies
from mDNS to be from the same address that was used for the request.

Jira: OSPRH-15638


beagles:
Draft as it depends on the config fix PR.

PR #367, opened 2025-08-21
Use quorum queues if enabled

lmiccini:
With https://github.com/openstack-k8s-operators/infra-operator/pull/429 we added the ability to configure RabbitMQ to use quorum queues.

This commit adds a check to verify if "quorumqueues" has been added to the TransportURL secret. If its value is true, the services will be configured to use quorum queues.

Jira: https://issues.redhat.com/browse/OSPRH-19160


tests/functional/base_test.go
```diff
@@ -78,6 +78,22 @@ func CreateTransportURLSecret(name types.NamespacedName) *corev1.Secret {
 	return secret
 }
 
+func CreateTransportURLSecretWithQuorumQueues(name types.NamespacedName, quorumQueues bool) *corev1.Secret {
```
fmount, line 81: @lmiccini should this be updated to point to the function provided by infra-operator?

stuggi (approved):
/lgtm

PR #440, opened 2026-03-23
Only add TLS related env if TLS enabled.

beagles:
Fixes some TLS flexibility issues
    
Sets up the pool jobs and the redis URL to work whether TLS is enabled or not (previously assumed tls was always enabled).
    
Also fixes a redis instance monitoring bug.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/api/bases/designate.openstack.org_designates.yaml
+++ b/api/bases/designate.openstack.org_designates.yaml
@@ -2381,6 +2381,9 @@ spec:
                 items:
                   type: string
                 type: array
+              redisTLS:
+                description: RedisTLS - whether the Redis instance has TLS enabled
+                type: string
               transportURLSecret:
                 description: TransportURLSecret - Secret containing RabbitMQ transportURL
                 type: string
--- a/config/crd/bases/designate.openstack.org_designates.yaml
+++ b/config/crd/bases/designate.openstack.org_designates.yaml
@@ -2381,6 +2381,9 @@ spec:
                 items:
                   type: string
                 type: array
+              redisTLS:
+                description: RedisTLS - whether the Redis instance has TLS enabled
+                type: string
               transportURLSecret:
                 description: TransportURLSecret - Secret containing RabbitMQ transportURL
                 type: string
--- a/internal/controller/designate_controller.go
+++ b/internal/controller/designate_controller.go
@@ -465,6 +465,30 @@ func (r *DesignateReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Man
 		return nil
 	}
 
+	// Watch for changes to the Redis CR used by Designate (e.g. TLS config changes)
+	redisWatchFn := func(_ context.Context, o client.Object) []reconcile.Request {
+		designates := &designatev1beta1.DesignateList{}
+		listOpts := []client.ListOption{
+			client.InNamespace(o.GetNamespace()),
+		}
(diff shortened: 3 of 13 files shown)


config/crd/bases/designate.openstack.org_designates.yaml
```diff
@@ -2381,6 +2381,9 @@ spec:
                 items:
                   type: string
                 type: array
+              redisTLS:
+                description: RedisTLS - whether the Redis instance has TLS enabled
+                type: string
```
omersch381, line 2386: Thinking out loud, do we want to store it as a string or boolean?

internal/controller/designate_controller.go
```diff
@@ -465,6 +465,30 @@ func (r *DesignateReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Man
 		return nil
 	}
 
+	// Watch for changes to the Redis CR used by Designate (e.g. TLS config changes)
+	redisWatchFn := func(_ context.Context, o client.Object) []reconcile.Request {
```
omersch381, line 469: nit: we might want to use ctx parameter instead of discarding it with _ and then using it in the r.List call, like we do in findDesignatesForMultipoolConfigMap

internal/controller/designate_controller.go
```diff
@@ -465,6 +465,30 @@ func (r *DesignateReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Man
...
+		designates := &designatev1beta1.DesignateList{}
+		listOpts := []client.ListOption{
+			client.InNamespace(o.GetNamespace()),
+		}
+		if err := r.List(context.Background(), designates, listOpts...); err != nil {
+			Log.Error(err, "Unable to retrieve Designate CRs %v")
```
omersch381: nit: Claude says:

""
Log.Error(err, "Unable to retrieve Designate CRs") is sufficient. The err is already printed. The %v in the current code is just a literal string that gets output as-is — it doesn't get substituted with anything.
""

And we have many instances of it in the code. If it's right, maybe we would like to change that in the future

config/crd/bases/designate.openstack.org_designates.yaml
beagles: There is a check against using booleans in the CR so new ones are to be avoided... apparently :)

internal/controller/designate_controller.go
beagles: Good point

beagles: there was a series of earlier commits that did something similar but used %w which a quick google suggests is better. Worth investigating. In the meantime, I'll remove it.

internal/controller/designate_controller.go
beagles: this is the pattern we've been using. It is worth considering but would be better served in a separate specific PR.

omersch381 (approved):

api/bases/designate.openstack.org_designates.yaml
```diff
@@ -2381,6 +2381,9 @@ spec:
                 items:
                   type: string
                 type: array
+              redisTLS:
```
stuggi, line 2384: @beagles @omersch381 why is this parameter needed when we get bellow if redis is configured for tls from redis?

@lmiccini fyi

PR #447, opened 2026-04-10
mdns setipalias: include cluster pod IP in listen addresses

beagles:
Extend setipalias.sh to listen on both the predictable NAD IP (when available) and the pod's cluster IP from /etc/hosts, building a comma-separated listen value for service:mdns. A format_listen_addr helper ensures IPv6 addresses are correctly bracketed per RFC 3986.

Resolves: [OSPRH-22265](https://redhat.atlassian.net/browse/OSPRH-22265)

### DIFF
--- a/templates/designatemdns/bin/setipalias.sh
+++ b/templates/designatemdns/bin/setipalias.sh
@@ -15,12 +15,45 @@
 # under the License.
 set -ex
 
-# expect that the common.sh is in the same dir as the calling script
-#
-SCRIPTPATH="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
 SVC_CFG_MERGED=/var/lib/config-data/merged/designate.conf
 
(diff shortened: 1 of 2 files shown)


templates/designatemdns/bin/setipalias.sh
```diff
@@ -15,12 +15,45 @@
...
+        echo "[${addr}]:${port}"
+    else
+        echo "${addr}:${port}"
+    fi
+}
+
+IPADDR=$(/usr/local/bin/container-scripts/setipalias.py) || true
+LISTEN_VALUE=""
+SEPARATOR=""
 if [ $? -eq 0 ] && [ -n "$IPADDR" ]; then
```
omersch381: We might want to remove the [ $? -eq 0 ] now, as we have || true before that line

beagles: Good point. I'll fix.

omersch381 (approved):

PR #450, opened 2026-04-15
Simplify Redis TLS handling using upstream status field

lmiccini:
Instead of manually querying the Redis service for IPs and checking
TLS configuration, use the GetRedisClientURL() method from the
infra-operator Redis API which returns the correct URL based on the
Redis status field.
    
The RedisTLS status field is removed as it was recently introduced
and not yet released downstream. The RedisHostIPs field in the API
types is kept but marked as deprecated to preserve API compatibility.


stuggi:
Thanks Luca, looks good to me, @beagles if you are good with it, too, we have to override pre-commit  to allow the crd change.

beagles:
> Thanks Luca, looks good to me, @beagles if you are good with it, too, we have to override pre-commit to allow the crd change.

Yes. This is a massive improvement IMO. Thanks!

stuggi (approved):

PR #469, opened 2026-05-25
Add predictable IP support for Unbound

abhiramnarayana:
Enable predictable IP allocation for Unbound pods, using the same mechanism as mDNS and BIND. Unbound IPs are allocated last in the chain (mDNS → BIND → Unbound) to avoid disrupting existing assignments on upgrades.

Adds NetUtilsImage field to DesignateUnboundSpecBase, a predictableips init container to the Unbound StatefulSet, and the designate-unbound-ip-map ConfigMap managed by the parent controller. Unbound does not need a designate.conf init container since it uses its own unbound.conf configuration.

### DIFF
--- a/internal/controller/designate_controller.go
+++ b/internal/controller/designate_controller.go
@@ -1172,12 +1172,22 @@ func (r *DesignateReconciler) reconcileNormal(ctx context.Context, instance *des
 		return ctrl.Result{}, err
 	}
 
+	// Handle Unbound predictable IPs configmap
+	unboundLabels := labels.GetLabels(instance, labels.GetGroupLabel(instance.Name), map[string]string{})
+	unboundConfigMap, err := r.handleConfigMap(ctx, helper, instance, designate.UnboundPredIPConfigMap, unboundLabels)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
 	allocatedIPs := make(map[string]bool)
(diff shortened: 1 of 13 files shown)


internal/controller/designate_controller.go
```diff
@@ -1232,6 +1232,34 @@ func (r *DesignateReconciler) reconcileNormal(ctx context.Context, instance *des
 		return ctrlResult, err
 	}
 
+	// Handle Unbound predictable IPs configmap
+	unboundLabels := labels.GetLabels(instance, labels.GetGroupLabel(instance.Name), map[string]string{})
+	unboundConfigMap, err := r.handleConfigMap(ctx, helper, instance, designate.UnboundPredIPConfigMap, unboundLabels)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
```
omersch381: Can we move those lines to be right after line 1176?

And then add 
```
for _, key := range slices.Sorted(maps.Keys(unboundConfigMap.Data)) {
		allocatedIPs[unboundConfigMap.Data[key]] = true
}
```

omersch381 (changes requested):

internal/controller/designate_controller.go
abhiramnarayana: Done.

omersch381 (approved):
LGTM

PR #470, opened 2026-05-29
Support external bind information (mounting rndc keys only)

beagles:
Step 1 of supporting external binds is to get the rndc keys in position so bind pool targets will work.

### DIFF
--- a/internal/designateworker/deployment.go
+++ b/internal/designateworker/deployment.go
@@ -46,17 +46,48 @@ func Deployment(
 	rootUser := int64(0)
 	serviceName := fmt.Sprintf("%s-worker", designate.ServiceName)
 
-	volumeDefs := append(designate.GetStandardVolumeMapping(instance),
-		designate.VolumeMapping{Name: designate.DesignateBindKeySecret, Type: designate.SecretMount, MountPath: "/etc/designate/rndc-keys"})
+	volumes, initVolumeMounts := designate.ProcessVolumes(designate.GetStandardVolumeMapping(instance))
+	projectedVolumes := []corev1.VolumeProjection{
+		{
+			Secret: &corev1.SecretProjection{
+				LocalObjectReference: corev1.LocalObjectReference{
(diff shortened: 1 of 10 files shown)


internal/designateworker/deployment.go
```diff
@@ -46,17 +46,46 @@ func Deployment(
...
+				},
+			},
+		},
+		{
+			Secret: &corev1.SecretProjection{
+				LocalObjectReference: corev1.LocalObjectReference{
+					Name: designate.ExternalRndcData,
+				},
+			},
+		},
```
omersch381, line 64: Should we add the Optional flag to those secrets in this PR instead of 471?

internal/designateworker/deployment.go
```diff
@@ -46,17 +46,46 @@ func Deployment(
...
+	}
+	projectedVolume := corev1.Volume{
+		Name: "worker-projected-volumes",
+		VolumeSource: corev1.VolumeSource{
+			Projected: &corev1.ProjectedVolumeSource{
+				Sources: projectedVolumes,
+			},
+		},
+	}
```
omersch381, line 75: Should we set its DefaultMode to 0640?

omersch381 (commented):
A few comments inline

internal/designateworker/deployment.go
beagles: We could. It may not be necessary at all, but I added afterwards as an afterthought "just-in-case". There is an argument that "optional" might in fact be wrong and would hide a bug (ie. the rndc secret is expected to exist when the pod template spec changes - because it happens after the rndc secret creation in the main reconcile loop). Let's discuss it and if we think it's a bad idea, I'll remove it from 471.

internal/designateworker/deployment.go
beagles: ooh yes, good catch

omersch381 (approved):

PR #474, opened 2026-06-09
Support configuring external access to DNS masters

beagles:
If mDNS services with the appropriate labelling are present, then
override the masters for external targets with those IPs.

### DIFF
--- a/internal/controller/designate_controller.go
+++ b/internal/controller/designate_controller.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"maps"
 	"slices"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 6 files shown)


beagles:
holding on merge of commit that is in a separate PR- I'll also move docs out to a separate PR.

internal/controller/designate_controller.go
```diff
@@ -503,6 +540,25 @@ func (r *DesignateReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Man
 		return result
 	}
 
+	externalMasterServiceFn := func(_ context.Context, o client.Object) []reconcile.Request {
+		if o.GetLabels()[designate.ExternalMasterServiceLabel] != "true" {
+			return nil
+		}
+
+		designates := &designatev1beta1.DesignateList{}
+		if err := r.List(context.Background(), designates, client.InNamespace(o.GetNamespace())); err != nil {
```
omersch381: Should we use ctx instead of context.Background()?

internal/controller/designate_controller.go
```diff
@@ -111,6 +112,41 @@ func getSecret(
...
+	}
+	ips := make([]string, 0, len(services.Items))
+	pendingIngress := false
+	for _, service := range services.Items {
+		if service.Spec.Type != corev1.ServiceTypeLoadBalancer {
+			continue
+		}
+		// If the service is missing the annotation, apply to all pools by skipping the pool name check.
+		servicePool, ok := service.Annotations[designate.ExternalPoolServiceAnnotation]
+		if ok && servicePool != poolName {
```
omersch381, line 135: Do we test the "ok == false" case? If not, do we want to?

internal/controller/designate_controller.go
beagles: you are correct! fixing

internal/controller/designate_controller.go
beagles: So what we are checking here is if a pool name is set and it doesn't match the pool name, then skip it. If it isn't set, e.g ok == false it falls through. This enables using one set of kubernetes services to apply to all external BINDs.. so I think this is correct.

omersch381 (approved):
LGTM

PR #475, opened 2026-06-10
Add AZ-aware mode CR field and extend ConfigMap validation

abhiramnarayana:
Add AZAwareMode string enum field (DesignateAZMode type) to DesignateSpecBase with allowed values "" and "Enabled", following OpenShift API conventions. Provides a declarative way to enable AZ-aware multipool with BIND views, per-pool TSIG keys, and per-AZ Unbound instances.

Extend validateMultipoolConfig to reject pools missing a view when AZ mode is enabled. Use variadic parameter for backward compatibility with existing callers. Regenerate CRD manifests and add unit tests for AZ-aware validation rules.

### DIFF
--- a/api/v1beta1/designate_types.go
+++ b/api/v1beta1/designate_types.go
@@ -26,7 +26,17 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// DesignateAZMode describes the AZ-aware multipool mode setting
+// +kubebuilder:validation:Enum="";Enabled;Disabled
+type DesignateAZMode string
+
 const (
+	// AZModeEnabled activates AZ-aware multipool with BIND views and per-pool TSIG keys
+	AZModeEnabled DesignateAZMode = "Enabled"
+
+	// AZModeDisabled explicitly disables AZ-aware multipool mode
+	AZModeDisabled DesignateAZMode = "Disabled"
+
 	// DbSyncHash hash
 	DbSyncHash = "dbsync"
 
@@ -235,6 +245,12 @@ type DesignateSpecBase struct {
 	// +kubebuilder:validation:Optional
 	// ExternalBindsSecret is the name of the secret containing external BIND9 configurations
 	ExternalBindsSecret string `json:"externalBindsSecret,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// AZAwareMode - when set to "Enabled", activates AZ-aware multipool mode with
+	// BIND views, per-pool TSIG keys, and per-AZ Unbound instances. Requires a valid
+	// multipool ConfigMap with view fields defined for each pool.
+	AZAwareMode DesignateAZMode `json:"azAwareMode,omitempty"`
 }
 
 // DesignateStatus defines the observed state of Designate
--- a/internal/controller/designate_controller.go
+++ b/internal/controller/designate_controller.go
@@ -1108,7 +1108,7 @@ func (r *DesignateReconciler) reconcileNormal(ctx context.Context, instance *des
 	}
 
 	// Get multipool configuration once for use in bind IP allocation and pools.yaml generation
-	multipoolConfig, err := designate.GetMultipoolConfig(ctx, helper.GetClient(), instance.Namespace)
+	multipoolConfig, err := designate.GetMultipoolConfig(ctx, helper.GetClient(), instance.Namespace, instance.Spec.AZAwareMode)
 	if err != nil {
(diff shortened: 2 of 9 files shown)


internal/designate/generate_bind9_pools_yaml.go
```diff
@@ -140,7 +142,7 @@ type MultipoolConfig struct {
 
 // GetMultipoolConfig reads and parses the multipool ConfigMap
 // Returns nil if ConfigMap doesn't exist
-func GetMultipoolConfig(ctx context.Context, k8sClient client.Client, namespace string) (*MultipoolConfig, error) {
+func GetMultipoolConfig(ctx context.Context, k8sClient client.Client, namespace string, azAwareMode ...bool) (*MultipoolConfig, error) {
```
omersch381: Why variadic bool instead of bool?
If we use bool, I think we could drop lines 176-179

internal/controller/designate_controller.go
```diff
@@ -972,7 +972,7 @@ func (r *DesignateReconciler) reconcileNormal(ctx context.Context, instance *des
 	Log.Info("Deployment API task reconciled")
 
 	// Get multipool configuration once for use in bind IP allocation and pools.yaml generation
-	multipoolConfig, err := designate.GetMultipoolConfig(ctx, helper.GetClient(), instance.Namespace)
+	multipoolConfig, err := designate.GetMultipoolConfig(ctx, helper.GetClient(), instance.Namespace, instance.Spec.AZAwareMode == designatev1beta1.AZModeEnabled)
```
omersch381: maybe just designatev1beta1.AZModeEnabled?

internal/designate/generate_bind9_pools_yaml.go
omersch381: Maybe we could also add a disabled mode, and change the condition to only take action if it is enabled, with some string manipulation, WDYT?

omersch381 (changes requested):
I added some comments inline.

internal/designate/generate_bind9_pools_yaml.go
abhiramnarayana: Changes are done.

internal/controller/designate_controller.go
abhiramnarayana: Done.

omersch381 (approved):
LGTM

api/v1beta1/designate_types.go
```diff
@@ -235,6 +245,12 @@ type DesignateSpecBase struct {
 	// +kubebuilder:validation:Optional
 	// ExternalBindsSecret is the name of the secret containing external BIND9 configurations
 	ExternalBindsSecret string `json:"externalBindsSecret,omitempty"`
+
+	// +kubebuilder:validation:Optional
+	// AZAwareMode - when set to "Enabled", activates AZ-aware multipool mode with
+	// BIND views, per-pool TSIG keys, and per-AZ Unbound instances. Requires a valid
+	// multipool ConfigMap with view fields defined for each pool.
+	AZAwareMode DesignateAZMode `json:"azAwareMode,omitempty"`
```
stuggi, line 253: @abhiramnarayana @omersch381 does this new parameter change any deployment spec? wondering if it wasn't enough to just pass a valid multipool configmap with view fields defined for each pool.

abhiramnarayana: hi @stuggi .. No, this doesn't change any deployment spec. The azAwareMode field is optional (omitempty) and defaults to empty string — existing deployments are completely unaffected. To be more specific 
Since azAwareMode is omitempty — if not specified in the CR, it's "" (empty string)
GetMultipoolConfig passes instance.Spec.AZAwareMode which is "" for any existing deployment. And Inside validateMultipoolConfig, the check if azAwareMode == designatev1.AZModeEnabled is false.
So the view validation is skipped entirely for existing deployments

Further As for "just passing a valid configmap with views" — for validation alone that would work, but this field will be the opt-in gate for future operator behavior changes (per-pool TSIG key creation, per-AZ Unbound deployment, BIND view ACL generation) that can't be inferred from the ConfigMap contents. This was a design decision as discussed with @omersch381 and @beagles 

@omersch381 Feel free to add any other comments if any?

stuggi: ack, thanks for the explanation. I just wanted to make sure we do not introduce parameters which would  just set a config in a way the user could also do by just pass in a custom config via the config customization interfaces. this is something should not do.

PR #476, opened 2026-06-11
Revert unbound health checks to pgrep

beagles:
The unbound-streamtcp check seems to timeout occasionally causing unbound pods to recycle.


omersch381 (approved):
LGTM

PR #479, opened 2026-06-15
Fix a bad stubzone bug for unbound

beagles:
The bind map can now contain things other than bind ip data, causing bad configuration files being generated.

### DIFF
--- a/internal/controller/designateunbound_controller.go
+++ b/internal/controller/designateunbound_controller.go
@@ -637,10 +637,12 @@ func (r *UnboundReconciler) generateServiceConfigMaps(
 			}
 			return err
 		}
-		bindIPs := make([]string, len(bindIPMap.Data))
+		bindIPs := make([]string, 0, len(bindIPMap.Data))
 		keyTmpl := "bind_address_%d"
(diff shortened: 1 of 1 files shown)


internal/controller/designateunbound_controller.go
```diff
@@ -637,10 +637,12 @@ func (r *UnboundReconciler) generateServiceConfigMaps(
...
 			return err
 		}
-		bindIPs := make([]string, len(bindIPMap.Data))
+		bindIPs := make([]string, 0, len(bindIPMap.Data))
 		keyTmpl := "bind_address_%d"
 		for i := 0; i < len(bindIPMap.Data); i++ {
-			bindIPs[i] = bindIPMap.Data[fmt.Sprintf(keyTmpl, i)]
+			if ip, ok := bindIPMap.Data[fmt.Sprintf(keyTmpl, i)]; ok {
+				bindIPs = append(bindIPs, ip)
+			}
```
omersch381, line 645: LGTM. I wonder, do we want to log anything in case bindIPs length is 0? (I guess we could reach that situation because of various factors that go wrong).

omersch381 (approved):

PR #480, opened 2026-06-18
Add unit tests for AZ-aware struct and schema changes

abhiramnarayana:
Add tests to generate_bind9_pools_yaml_test.go verifying the data model and validation changes for AZ-aware multipool support protect against regressions. Covers view propagation to generated pool targets, template rendering of view and tsigkey_id fields (both inclusion when set and omission when empty), and AZ mode validation across all three enum states ("", "Disabled", "Enabled").

### DIFF
--- a/internal/designate/generate_bind9_pools_yaml_test.go
+++ b/internal/designate/generate_bind9_pools_yaml_test.go
@@ -19,6 +19,7 @@ package designate
 import (
 	"strings"
 	"testing"
+	"text/template"
 
 	designatev1 "github.com/openstack-k8s-operators/designate-operator/api/v1beta1"
 )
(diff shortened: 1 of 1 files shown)


internal/designate/generate_bind9_pools_yaml_test.go
```diff
@@ -756,3 +757,321 @@ func TestGenerateDefaultPoolWithExternalBindsFromYAML(t *testing.T) {
 		t.Errorf("expected external RNDC port %d, got %d", RNDCPort, externalTarget.Options.RNDCPort)
 	}
 }
+
+func TestGenerateMultiplePoolsWithViewAndTSIG(t *testing.T) {
```
omersch381: This test does not include tsig keys.

omersch381 (commented):
I left a comment regarding one of the tests name, but except for that, it LGTM

internal/designate/generate_bind9_pools_yaml_test.go
abhiramnarayana: Done.

abhiramnarayana:
recheck

omersch381 (approved):
LGTM

PR #485, opened 2026-06-23
18.0-fr6 Support overriding masters with mdns services.

beagles:
If mDNS services with the appropriate labelling are present, then override the masters for external targets with those IPs.


omersch381 (approved):
LGTM

PR #486, opened 2026-06-25
Add per-pool TSIG key creation gated by AZAwareMode

abhiramnarayana:
Replace the shared TSIG key model with per-pool keys when AZAwareMode=Enabled. Each pool (including default) gets its own TSIG key with scope=POOL and resource_id set to the actual pool UUID from Designate.

Implement the two-pass creation flow: pools are created without tsigkey_id, pool UUIDs are retrieved, TSIG keys are created per pool, then pools.yaml is regenerated with tsigkey_id populated.

Add lifecycle management: cleanup orphaned TSIG keys when pools are removed, clean transition between per-pool and shared modes. The shared TSIG model remains unchanged when AZAwareMode is not enabled.

Add unit tests for pool generation with TSIG key IDs.

### DIFF
--- a/internal/controller/designate_controller.go
+++ b/internal/controller/designate_controller.go
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"maps"
@@ -1337,7 +1338,26 @@ func (r *DesignateReconciler) reconcileNormal(ctx context.Context, instance *des
 			Data: make(map[string]string),
 		}
 
-		poolsYaml, poolsYamlHash, err := designate.GeneratePoolsYamlDataAndHash(updatedBindMap, mdnsConfigMap.Data, nsRecords, multipoolConfig, externalBindData, externalMasterServiceIPs)
+		// Read per-pool TSIG key ID mapping if AZ-aware mode is enabled
+		var tsigKeyIDs map[string]string
+		if instance.Spec.AZAwareMode == designatev1beta1.AZModeEnabled && multipoolConfig != nil {
+			tsigSecretName := fmt.Sprintf("%s-backendbind9%s", instance.Name, designate.TsigSecretSuffix)
+			tsigSecret := &corev1.Secret{}
+			err := helper.GetClient().Get(ctx, types.NamespacedName{Name: tsigSecretName, Namespace: instance.GetNamespace()}, tsigSecret)
+			if err != nil && !k8s_errors.IsNotFound(err) {
+				return ctrl.Result{}, fmt.Errorf("failed to get TSIG secret %s: %w", tsigSecretName, err)
+			}
+			if err == nil {
+				if data, exists := tsigSecret.Data[designate.TSIGKeyIDsDataKey]; exists {
+					if jsonErr := json.Unmarshal(data, &tsigKeyIDs); jsonErr != nil {
+						Log.Info(fmt.Sprintf("Failed to parse TSIG key ID mapping from secret, pools.yaml will be generated without tsigkey_ids: %v", jsonErr))
+					}
+				}
+			}
+			// If secret doesn't exist yet, tsigKeyIDs stays nil — pools.yaml is generated without tsigkey_ids (first pass)
+		}
+
+		poolsYaml, poolsYamlHash, err := designate.GeneratePoolsYamlDataAndHash(updatedBindMap, mdnsConfigMap.Data, nsRecords, multipoolConfig, externalBindData, externalMasterServiceIPs, tsigKeyIDs)
 		if err != nil {
 			return ctrl.Result{}, err
 		}
--- a/internal/controller/designatebackendbind9_multipool.go
+++ b/internal/controller/designatebackendbind9_multipool.go
@@ -24,6 +24,7 @@ import (
 	"context"
 	"crypto/rand"
 	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"maps"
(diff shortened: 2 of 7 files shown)


internal/designate/tsigkeys.go
```diff
@@ -185,8 +156,8 @@ func CreateTSIGKey(
 		return nil, fmt.Errorf("failed to create TSIG key: %w", err)
 	}
 
-	// Extract TSIGKey fields from response
 	tsigKey := &TSIGKey{
+		ID:        result["id"].(string),
 		Name:      result["name"].(string),
 		Algorithm: result["algorithm"].(string),
 		Secret:    result["secret"].(string),
```
omersch381: I know you only added the ID field, but I wonder, if any field is missing, or nil for any reason, the operator process panics. Should we add a comma, ok here?

internal/controller/designatebackendbind9_multipool.go
```diff
@@ -734,18 +735,153 @@ func (r *DesignateBackendbind9Reconciler) cleanupOrphanedStatefulSets(
...
 ) (ctrl.Result, error) {
 	Log := r.GetLogger(ctx)
-	Log.Info("Reconciling TSIG secret for multipool")
+
+	// Determine AZ-aware mode from parent Designate CR
+	designateInstance, err := r.getDesignateCR(ctx, helper, instance)
+	if err != nil {
+		// If we can't get the parent CR, fall back to shared mode
+		Log.Info(fmt.Sprintf("Could not get Designate CR for AZ mode check, using shared TSIG model: %v", err))
+		return r.reconcileSharedTSIGSecret(ctx, instance, helper, multipoolConfig)
```
omersch381: should we fall back to shared mode if we can't get the parent CR?

internal/controller/designate_controller.go
```diff
@@ -1337,7 +1338,23 @@ func (r *DesignateReconciler) reconcileNormal(ctx context.Context, instance *des
...
+			tsigSecret := &corev1.Secret{}
+			err := helper.GetClient().Get(ctx, types.NamespacedName{Name: tsigSecretName, Namespace: instance.GetNamespace()}, tsigSecret)
+			if err == nil {
+				if data, exists := tsigSecret.Data[designate.TSIGKeyIDsDataKey]; exists {
+					if jsonErr := json.Unmarshal(data, &tsigKeyIDs); jsonErr != nil {
+						Log.Info(fmt.Sprintf("Failed to parse TSIG key ID mapping from secret, pools.yaml will be generated without tsigkey_ids: %v", jsonErr))
+					}
+				}
+			}
+			// If secret doesn't exist yet, tsigKeyIDs stays nil — pools.yaml is generated without tsigkey_ids (first pass)
```
omersch381: and if we have another error? (say, except for secret not found)

internal/controller/designatebackendbind9_multipool.go
```diff
@@ -734,18 +735,153 @@ func (r *DesignateBackendbind9Reconciler) cleanupOrphanedStatefulSets(
...
+		}
+	} else if !k8s_errors.IsNotFound(err) {
+		return ctrl.Result{}, err
+	}
+
+	// Create per-pool TSIG keys in Designate
+	tsigKeys, err := r.ensurePerPoolTSIGKeys(ctx, helper, instance, multipoolConfig)
+	if err != nil {
+		if strings.Contains(err.Error(), "not ready") || strings.Contains(err.Error(), "no running") ||
+			strings.Contains(err.Error(), "not yet registered") {
```
omersch381: I think the only error we could actually get here is an error that would contain "not ready".

So we might also want to change the existing pattern at line ~974. Maybe we should switch to errors.Is()

internal/controller/designatebackendbind9_multipool.go
```diff
@@ -967,6 +1267,24 @@ func (r *DesignateBackendbind9Reconciler) getPoolConfigHash(multipoolConfig *des
 	return hash, nil
 }
 
+// getPerPoolConfigHash generates a hash including ALL pools (for per-pool TSIG model).
+func (r *DesignateBackendbind9Reconciler) getPerPoolConfigHash(multipoolConfig *designate.MultipoolConfig) (string, error) {
```
omersch381: We should also add mDNS IPs as change signal - their IPs are stored as "server" blocks in the bind config files. I.e. if their IPs are modified due to any change, the binds config secret won't be re-written.

internal/controller/designatebackendbind9_multipool.go
```diff
@@ -920,6 +1064,162 @@ func (r *DesignateBackendbind9Reconciler) ensureSharedTSIGKey(
...
+	sort.Strings(poolNames)
+
+	// Add key definitions for each pool
+	for _, poolName := range poolNames {
+		key := tsigKeys[poolName]
+		fmt.Fprintf(&config, "key \"%s\" {\n", key.Name)
+		fmt.Fprintf(&config, "    algorithm %s;\n", key.Algorithm)
+		fmt.Fprintf(&config, "    secret \"%s\";\n", key.Secret)
+		config.WriteString("};\n\n")
+	}
```
omersch381, line 1217: nit: we might want to validate the key names before we inject them directly to the bind config files.

internal/controller/designatebackendbind9_multipool.go
```diff
@@ -734,18 +735,153 @@ func (r *DesignateBackendbind9Reconciler) cleanupOrphanedStatefulSets(
...
+	if designateInstance.Spec.AZAwareMode == designatev1beta1.AZModeEnabled {
+		return r.reconcilePerPoolTSIGSecrets(ctx, instance, helper, multipoolConfig)
+	}
+
+	return r.reconcileSharedTSIGSecret(ctx, instance, helper, multipoolConfig)
+}
+
+// reconcilePerPoolTSIGSecrets creates per-pool TSIG keys and stores them in a Secret.
+// Every pool (including default) gets its own TSIG key with scope=POOL and resource_id=pool-UUID.
+func (r *DesignateBackendbind9Reconciler) reconcilePerPoolTSIGSecrets(
```
omersch381, line 762: Can we extract sub-functions from this function? It is very long IMO

omersch381 (changes requested):
A few comments inline

omersch381:
There are some parts in the code that allow / hint that transition from AZ aware env to non AZ aware (say, removing existing views) is allowed. In such a transition / removal, we'd lose replications, hence we lose data and therefore we shouldn't allow it. I will open another ticket that will add a webhook that catches flipping the AZ aware flag and does nothing when it happens.

internal/controller/designatebackendbind9_multipool.go
beagles, line 1: Some of (maybe most?) of these changes are not directly related to the configuration  and management of the BIND9 backend. We should revisit and refactor out of here. Note also that while the file is a "multipool" specific file, these are all bound to the DesignateBackendbind9Reconciler type which might lead to some confusing mistakes in the future.

internal/controller/designatebackendbind9_multipool.go
```diff
@@ -920,13 +1088,177 @@ func (r *DesignateBackendbind9Reconciler) ensureSharedTSIGKey(
...
+	for i := range existingKeys {
+		existingKeysByName[existingKeys[i].Name] = &existingKeys[i]
+	}
+
+	result := make(map[string]*designate.TSIGKey)
+
+	for _, pool := range multipoolConfig.Pools {
+		keyName := instance.Name + "-" + pool.Name + designate.PerPoolTSIGKeySuffix
+
+		poolUUID, exists := poolNameToID[pool.Name]
```
beagles, line 1117: We might want to look at how this works... IIUC, we create a have a deployment change around a pool, and we get this far without having designate "knowing about it". Is there a way to structure things that TSIG key and related items get managed with a view to the pool as it *will be* created and then do the following changes - the presumption being that the pool manage will eventually create the pools in designate. The question is I suppose what the harm is in pushing configuration in to BIND9 that isn't used. Maybe this will be more clear if we move this out of the BIND9 backend files and work it into the logic closer to where the pools yaml is generated?

internal/controller/designatebackendbind9_multipool.go
```diff
@@ -920,13 +1088,177 @@ func (r *DesignateBackendbind9Reconciler) ensureSharedTSIGKey(
...
+	}
+
+	// Add server blocks for each mdns IP with all keys
+	for _, mdnsIP := range mdnsIPs {
+		fmt.Fprintf(&config, "server %s {\n", mdnsIP)
+		config.WriteString("    keys {\n")
+		for _, poolName := range poolNames {
+			key := tsigKeys[poolName]
+			fmt.Fprintf(&config, "        %s;\n", key.Name)
+		}
```
beagles, line 1226: bind configs are getting complicated :) we should consider staging configuration like create a "proposed-bind9-config" and run a named-checkconf job on it and if successful, move the contents from proposed to whatever the runtime secret is called.

internal/controller/designatebackendbind9_multipool.go
```diff
@@ -957,8 +1289,26 @@ func (r *DesignateBackendbind9Reconciler) getPoolConfigHash(multipoolConfig *des
...
+// getPerPoolConfigHash generates a hash including ALL pools and mDNS IPs (for per-pool TSIG model).
+func (r *DesignateBackendbind9Reconciler) getPerPoolConfigHash(multipoolConfig *designate.MultipoolConfig, mdnsIPs []string) (string, error) {
+	var poolNames []string
+	for _, pool := range multipoolConfig.Pools {
+		poolNames = append(poolNames, pool.Name)
+	}
+
+	sort.Strings(poolNames)
+
+	hashInput := "per-pool:" + strings.Join(poolNames, ",") + "|mdns:" + strings.Join(mdnsIPs, ",")
```
beagles: A lot of duplication between getPerPoolConfigHash and getPoolConfigHash. Why not add a helper with a prefix of "" or "per-pool:"?

beagles (dismissed):
The get?PoolConfigHash functions are almost entirely duplicate code. I would like to see the internal logic collapsed into a single helper as suggested and both functions implemented in terms of that.

As mentioned in the file comment, I feel we need to revisit designatebackendbind9_multipool.go and refactor the bits that have nothing to do with BIND9 management or are heavily intertwined with other operator managed resources. The introduction of the TSIG key management and mDNS dependencies have made this pretty unweildy. We might also rename the file to bind9_multipoolextensions.go  as it is not a complete controller but additional implementationadded to the controller implemented in designate_backendbind9.go. This is bigger change and can be done after this. I am okay with merging this after the small refactor above is implemented, with the provision we set a task to start refactoring this asap.

internal/controller/designatebackendbind9_multipool.go
abhiramnarayana: Done

beagles:
thanks for fixing up the hash computation duplication

omersch381 (approved):
Eventually I have implemented the changes I was asking for. The current version of the change LGTM

PR #494, opened 2026-07-01
Use native gophercloud tsigkeys package instead of raw HTTP calls

omersch381:
Gophercloud v2.12.0 added openstack/dns/v2/tsigkeys support, so we can replace our custom raw HTTP implementation with the upstream package. This removes the custom TSIGKey struct, CreateTSIGKeyOpts, and the ListAllTSIGKeys/ListAllTSIGKeysWithID helpers, using gophercloud's List, Create, and Delete operations with proper server-side filtering.


omersch381:
CI failures are all the same root cause: gophercloud v2.12.0 (the first version with dns/v2/tsigkeys) requires go 1.25.0 in its go.mod. The Prow CI image (golang-1.24-sdk) runs Go 1.24.13, which makes go mod tidy refuse to resolve a dependency that declares a higher Go version. The code builds fine with Go 1.24 — it's only the tidy check that blocks.

This PR is parked until the CI image is upgraded to Go 1.25+. No code changes needed on our side once that happens.

PR #495, opened 2026-07-08
Fix per-pool TSIG: one Secret per pool instead of one merged Secret

omersch381:
In AZ-aware mode every pool gets its own TSIG key, but the per-pool TSIG code merged every non-default pool's key into a single Secret mounted identically into every non-default pool's StatefulSet. BIND's `server` statement only accepts one key per remote address, so as soon as 2+ non-default pools shared the mdns server IP (the normal AZ-aware case), named failed to parse tsigkeys.conf and the pool's pod crash-looped.

Each pool already runs as a separate BIND process, so it never needs another pool's key. Give each pool its own Secret containing only its own key/server block instead of merging them, reusing the existing single-key generateTSIGConfig helper. StatefulSet() now derives the TSIG secret name from its own statefulSetName instead of the shared instance name, so each pool naturally gets a distinct secret with no API changes. The pool0-named Secret keeps carrying the full pool-name -> tsigkey-id map that designate_controller.go reads for pools.yaml, so that path is untouched. Also clean up orphaned per-pool Secrets when a pool is removed.

Verified on a live cluster: a pool that previously crash-looped on
`/etc/named/tsigkeys.conf: '}' expected` now starts and stays Ready.

### DIFF
--- a/internal/controller/designatebackendbind9_multipool.go
+++ b/internal/controller/designatebackendbind9_multipool.go
@@ -387,21 +387,6 @@ func (r *DesignateBackendbind9Reconciler) reconcileMultipoolStatefulSets(
 	// Track expected StatefulSet names for cleanup
 	expectedStatefulSets := make(map[string]bool)
 
-	// Get TSIG secret resourceVersion to include in pod annotations for automatic restarts
-	// This ensures non-default pool pods restart when TSIG keys change
-	var tsigSecretResourceVersion string
-	tsigSecretName := instance.Name + designate.TsigSecretSuffix
-	tsigSecret := &corev1.Secret{}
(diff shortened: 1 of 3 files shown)


internal/controller/designatebackendbind9_multipool.go
beagles, line 1: Where this is server clause thing, is this a multipool thing or a view thing or -if multipool is it limited to multipool with shared DNS backends? If so, it might be good to distinguish it as such.

beagles: Never mind - sorted out the answer.

beagles (approved):
/lgtm

PR #496, opened 2026-07-08
Mount TSIG secret on pool0 in AZ-aware mode

omersch381:
Mount TSIG secret on pool0 in AZ-aware mode

In AZ-aware mode every pool (including the default) has its own
per-pool TSIG key, but the StatefulSet for pool0 was not mounting the
TSIG secret and its pods were not restarted on TSIG changes.

Move the TSIG inclusion decision from the StatefulSet builder (which
was inferring it from the StatefulSet name) to the caller, which
already knows the pool index and AZ mode. The StatefulSet function now
takes an explicit includeTSIG parameter again, but keeps deriving the
secret name from the pool's own StatefulSet name (via
tsigSecretNameForPool) rather than the old shared instance name, so
pool0 gets its own correctly-scoped Secret instead of a merged one.

Verified on a live cluster: pool0 now mounts its own TSIG secret
alongside pool1's separate secret, and both pods stay Ready.

Depends-On: https://github.com/openstack-k8s-operators/designate-operator/pull/495

### DIFF
--- a/internal/controller/designatebackendbind9_controller.go
+++ b/internal/controller/designatebackendbind9_controller.go
@@ -646,7 +646,7 @@ func (r *DesignateBackendbind9Reconciler) reconcileSingleStatefulSet(
 
 	// Define a new StatefulSet object
 	// Use default bind IP ConfigMap for single-pool mode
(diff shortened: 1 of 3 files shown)


beagles (approved):
/lgtm

PR #505, opened 2026-07-24
Remove unnecessary pods RBAC permissions

stuggi:
The workload rbacRules and a kubebuilder RBAC marker granted full CRUD (create/delete/get/list/patch/update/watch) on core Pods, which was never exercised. Remove the unused permission and regenerate config/rbac/role.yaml.

The narrower get;list markers on the sub-controllers, and the pods/log marker, are kept, since those controllers watch worker pods and fetch pod logs for DNS pool status checks.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -8,7 +8,6 @@ rules:
   - ""
   resources:
   - configmaps
-  - pods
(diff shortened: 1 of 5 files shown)


abays (approved):
/lgtm

PR #515, opened 2026-07-29
Drop unnecessary capabilities from predictableips init container

omersch381:
setipalias.py only needs NET_ADMIN for the netlink address add. SYS_ADMIN and SYS_NICE were unused and, combined with the CR-supplied NetUtilsImage being untrusted, widened the container-escape surface under the privileged SCC.

Drop ALL and add back only NET_ADMIN plus CHOWN (needed by mdns, which also rewrites designate.conf via crudini and must preserve its root:designate group). Also set allowPrivilegeEscalation: false and a RuntimeDefault seccomp profile.


beagles (approved):
/lgtm

PR #524, opened 2026-08-06
Remove kolla entrypoint pattern from Designate, harden SecurityContext

stuggi:
Eliminate kolla_start/kolla_set_configs and the config.json staging pattern for designate-api, -central, -worker, -producer, and -mdns. The crudini-based config-merge init container (parent + child config secrets merged into a config-data-merged emptyDir) is kept exactly as-is -- only kolla's staging-to-final copy step is replaced with SubPath mounts of the same emptyDir at final paths.

Key changes:
- Replace kolla_start with the real per-service command (designate-central/-worker/-producer/-mdns --config-file ... --config-dir ...; /usr/sbin/httpd -DFOREGROUND for designate-api)
- designate-producer and designate-mdns previously had no explicit Command at all -- they silently relied on the image's kolla-based default ENTRYPOINT; both needed a Command added, not just replaced
- db-sync: exec bootstrap.sh directly instead of through kolla_start
- Mount designate.conf/custom.conf/my.cnf via SubPath from the merged emptyDir to their final paths; designate-api additionally mounts httpd.conf/ssl.conf the same way, plus new run-httpd/var-log-httpd emptyDirs
- Add GetConfigOverwriteVolumeMounts() to mount DefaultConfigOverwrite keys (e.g. policy.yaml) individually, since they are not unconditionally present
- Remove all KOLLA_CONFIG_STRATEGY env vars and delete the 6 now unreferenced *-config.json templates (api, central, worker, producer, mdns, db-sync)
- Use pod.RestrictiveSecurityContext/RestrictivePodSecurityContext from lib-common (serviceuser.DesignateUID) on all 5 migrated services plus the pool_update.go Jobs (PoolUpdateJob, PoolListJob)
- Init containers (SimpleInitContainer, InitContainer) changed from RunAsUser: 0 to RestrictiveSecurityContext(DesignateUID) -- init.sh only does cp/crudini into an emptyDir, no root needed
- init.sh: cp -a changed to cp -r to avoid attempting ownership preservation that silently fails as non-root
- Fix the TLS cert staging-path bug on designate-api: the per-endpoint TLS loop had no CertMount/KeyMount override, so certs landed at lib-common's default staging path while the rendered httpd.conf's SSLCertificateFile/SSLCertificateKeyFile already pointed at /etc/pki/tls/{certs,private}/<endpt>.{crt,key}
- httpd.conf: User/Group apache -> designate
- RBAC: add nonroot-v2 additively (anyuid;privileged;nonroot-v2) in both the kubebuilder marker and the runtime rbacRules slice
- DefaultMode 0440 for all config/secret volume mounts (ProcessVolumes modeMap, backendbind9 volumes, worker projected vol, pool_update PoolUpdateJob/PoolListJob config volumes)
- AutomountServiceAccountToken: ptr.To(false) added to all workloads (api, central, mdns, producer, worker, unbound, backendbind9, db-sync job, pool-update/pool-list jobs)
- designate-unbound SecurityContext hardened: added pod-level RestrictivePodSecurityContext(DesignateUID) with per-container RunAsUser: 0 + RunAsNonRoot: false override (ironic dnsmasq pattern) -- unbound must start as root to bind port 53
- backendbind9 inline emptyDir volumes replaced with lib-common volume.WritableDirVolume() helpers
- Removed stale "Keeps kolla happy" comment from init.sh
- Update functional and kuttl tests

The predictableips init container keeps root + NET_ADMIN/CHOWN for multus network interface setup (netlink address add) and crudini config ownership preservation. SYS_ADMIN/SYS_NICE were removed as they are unused by setipalias.py.

designate-backend-bind9 and designate-unbound keep RunAsUser 0 unchanged: both self-drop privileges after binding port 53.

designate-backend-bind9's own kolla removal is deliberately deferred: its config generation runs through a bespoke per-pod init.sh that selects an RNDC/TSIG key by StatefulSet pod ordinal, does a whole-directory copy into /etc/named, and relies on kolla's own "permissions" block -- meaningfully more complex than every other service's shared merge script, and since bind9 stays root anyway (no hardening benefit), the risk outweighed the consistency benefit. Its KOLLA_CONFIG_STRATEGY env var was deliberately left in place: removing just the env var while keeping kolla_start would silently flip its copy strategy from COPY_ALWAYS to kolla's default COPY_ONCE.

Jira: [OSPRH-33504](https://redhat.atlassian.net/browse/OSPRH-33504)
Jira: [OSPRH-33503](https://redhat.atlassian.net/browse/OSPRH-33503)

Depends-On: openstack-k8s-operators/lib-common#728


internal/designatebackendbind9/deployment.go
abays, line 242: I think `SimpleInitContainer` does a full replace of `SecurityContext`, so it drops `AllowPrivilegeEscalation`, `SeccompProfile`, `Capabilities.Drop`, and `RunAsNonRoot`.  Is this intentional?

stuggi: yeah thats to much, although its only the bind9 one which passes an sc and therefore would drop those.  I did change e.g. the mdns init container to not run as root. I think this could also work with the bind9. let me look into that

abays (approved):
/lgtm

